PR #233, opened 2026-05-13
fix(file_watcher): reset stop event on restart

aqilaziz:
## Summary
- reset the file watcher stop event when starting after a previous close
- avoid clearing the file store when no file store is configured
- make BaseFileWatcher tests use async-compatible file store mocks and cover restart behavior

## Context
This follows the watcher lifecycle investigation from agentscope-ai/QwenPaw#4224. Without resetting the previous stop signal, a restarted watcher can report as running while the watch loop exits immediately.

## Tests
- .\\.venv\\Scripts\\python.exe -m pytest tests\\test_base_file_watcher.py -q
- .\\.venv\\Scripts\\python.exe -m pre_commit run --files reme\\core\\file_watcher\\base_file_watcher.py tests\\test_base_file_watcher.py

### DIFF
--- a/reme/core/file_watcher/base_file_watcher.py
+++ b/reme/core/file_watcher/base_file_watcher.py
@@ -76,12 +76,14 @@ class BaseFileWatcher:
         if self._running:
             return
 
+        self._stop_event = asyncio.Event()
         self._running = True
 
         async def _initialize_and_watch():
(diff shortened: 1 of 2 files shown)


jinliyl (approved):
LGTM

PR #246, opened 2026-05-19
fix(core): correct tool results directory naming

jinliyl:
- Changed directory name from 'tool_result' to 'tool_results' in documentation
- Updated path variable assignment to use correct plural form 'tool_results'
- Ensured consistent directory naming throughout initialization logic


ployts (approved):
LGTM

PR #253, opened 2026-05-22
feat(file_store): add FAISS-backed local file store implementation

jinliyl:
- Introduce FaissLocalFileStore class with vector search capabilities using FAISS IndexFlatIP
- Implement FAISS index persistence with binary format and JSON id-map sidecar
- Add automatic index rebuilding when sidecar files are missing or corrupted
- Support tombstone mechanism for efficient deletion and compaction
- Register 'faiss' component type in the registry system
- Add faiss-cpu dependency requirement to pyproject.toml
- Update configuration schema to use simplified parameter structure
- Enhance search step to support parameter override from runtime context
- Add comprehensive unit tests for FAISS store functionality
- Implement fallback to parent methods for basic CRUD operations

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,6 +75,10 @@ ray = [
     "ray",
 ]
 
+faiss = [
+    "faiss-cpu>=1.7.4",
+]
+
 dev = [
     "jupyter-book",
     "ghp-import",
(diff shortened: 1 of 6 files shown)


imrewce (approved):

imrewce:
lgtm

PR #258, opened 2026-05-25
refactor(reme4): restructure steps packages

jinliyl:
- split steps into `crud` (vault-only ops), `transfer` (vault ↔ local fs bridges: upload/download/upload_resource), and `index` (search/reindex/update_catalog/update_index) packages
  - merge `graph_traverse` into `common.traverse`; support multi-seed BFS and build adjacency via `file_store.get_nodes()`
  - replace background `update_store`+`index_changes` with `scan_changes` + configurable `dispatch_job` dispatch pattern
  - introduce `file_catalog` component: JSONL-persisted FileNode catalog keyed by path
  - rewrite BM25 inverted index on numpy arrays with lazy deletion, doc_idx mapping, and tokenizer/version-namespaced on-disk pickle
  - factor `Application.__init__` into typed `_instantiate` helper; switch `jobs` config from list to dict; add `component_metadata_path` on `BaseComponent` and split dependency resolution into per-mode helpers
  - filter non-servable jobs from `help_step` via `enable_serve`
  - restructure tests: drop `test_bm25_lite.py`, add `test_keyword_index.py` and `test_file_catalog.py`; rename background-step tests for the new naming


ployts (approved):
LGTM

PR #265, opened 2026-05-29
refactor(evolve): replace ReActAgent with FlexReActAgent to allow structured output

jinliyl:
- Create FlexReActAgent subclass that overrides _reasoning method to handle tool_choice parameter
- Modify auto_memory_planner to use FlexReActAgent instead of ReActAgent
- Update base_step.py to accept additional kwargs in add_as_tool method
- Change run_job function to merge kwargs properly when calling jobs
- Remove redundant imports and constants from file_io.py
- Simplify _render_notes_block and rename _replace_or_append_notes to _rebuild_body
- Update method calls to use new function names in file_io operations

### DIFF
--- a/reme4/steps/base_step.py
+++ b/reme4/steps/base_step.py
@@ -188,14 +188,14 @@ class BaseStep(ABC):
             raise RuntimeError(f"Job {name} not found")
         return await job(**kwargs)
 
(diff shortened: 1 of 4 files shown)


ployts (approved):
LGTM

PR #269, opened 2026-06-01
feat(base_step): set default language from app context when not provided

jinliyl:
- Initialize language attribute with app context language if not explicitly set
- Add conditional logic to check for existing language value before assignment
- Ensure proper fallback behavior when language parameter is empty or None


ployts (approved):
LGTM

PR #270, opened 2026-06-02
feat(application): add thread pool support and background job threading capabilities

jinliyl:
- Integrate ThreadPoolExecutor for shared thread pool management in application context
- Add thread_pool_max_workers configuration option with default value of 0 (disabled)
- Implement thread pool creation and shutdown in application lifecycle methods
- Add use_thread_pool option to background job configuration for thread-based execution
- Support both asyncio event loop and threading event for background job stop mechanism
- Implement _run_in_thread method for running supervisors in dedicated threads
- Modify embedding models to use serial batching instead of concurrent batching
- Remove max_concurrency parameter from base embedding model component
- Update embedding model documentation to reflect serial batching implementation
- Add pyproject.toml with project metadata, dependencies, and build configuration

### DIFF
--- a/reme4/application.py
+++ b/reme4/application.py
@@ -2,6 +2,7 @@
 
 import asyncio
 import heapq
+from concurrent.futures import ThreadPoolExecutor
 from pathlib import Path
 from typing import AsyncGenerator, TypeVar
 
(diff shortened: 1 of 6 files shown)


ployts (approved):
LGTM

PR #292, opened 2026-06-24
feat(file_store): add embedding backfill for persisted chunks

jinliyl:
- Implement _backfill_missing_embeddings method to handle chunks without embeddings
- Add logic to identify and process chunks that predate embedding feature
- Integrate backfill process into store loading sequence
- Add proper error handling and logging for backfill operations
- Create unit test for embedding backfill functionality
- Ensure backfilled embeddings are properly persisted to storage


ployts (approved):
LGTM

Issue #302, opened 2026-06-29
ReMeLight LoCoMo Eval: Multi Hop stuck at ~53% despite semantic chunking + multi-query — seeking guidance


iluv7:
Hi, I've been benchmarking ReMeLight (file-based) on LoCoMo and wanted to share findings + ask for advice on the Multi Hop bottleneck.

### Current Results (semantic chunking + time fix)

| Metric      | Score  |
|-------------|--------|
| Overall     | 70.82% |
| Single Hop  | 72.09% |
| **Multi Hop** | **53.97%** |
| Temporal    | 92.31% |
| Open Domain | 77.19% |

The first four look reasonable. Multi Hop I can't get past 53%.

### What I've Tried

**1. Semantic Chunking**

Replaced the default recursive 400-token splitter with embedding-similarity-based breakpoints (bottom 25th percentile cosine between adjacent lines as adaptive threshold). Single Hop jumped 14pp; Multi Hop barely moved (+1.6pp).

**2. Fixed Summarizer Timestamps**

The summarizer was using `datetime.now()` for memory file names, so LoCoMo's 2023 conversations were written as `memory/2026-06-29.md`, invalidating all time-related answers. After extracting `time_created` from message metadata, Multi Hop went from 7.94% → 52.38%.

**3. LLM Multi-Query Decomposition + RRF Fusion**

Synchronous LLM call to generate 4 semantic query variants per search, concurrent vector + keyword per variant, cross-query RRF merge (original query full weight, variants decayed by 30).

### Remaining Errors

All post-53% errors are LLM temporal reasoning failures. Even with correct dates in file names and content, the LLM consistently conflates "when the conversation happened" with "when the event happened", or anchors "recently" to the current date:

- Q: "When did Jon lose his job?" Gold: "19 January, 2023" → answered: "28 June, 2026"
- Q: "When did Caroline attend the support group?" Gold: "7 May, 2023" → answered: "29 June, 2026"

### Questions

1. Are there known directions for improving Multi Hop in the file-based architecture beyond what I've tried?
2. Any recommended prompt techniques to force temporal context preservation during `summary_memory`?
3. In the vector-based version, how are temporal facts mapped to `when_to_use` vs `content`? Curious about the architectural difference in temporal reasoning between the two versions.

jinliyl:
[...]

PR #305, opened 2026-06-30
fix(auto-memory): allow explicit daily dates

RerankerGuo:
## Summary
- add an optional `date` parameter to `auto_memory` and `daily_write` for historical daily-note targets
- validate explicit daily dates as `YYYY-MM-DD` while preserving the runtime-date default
- document historical conversation usage and add regression coverage

Refs #302

## Verification
- `UV_CACHE_DIR=/tmp/uv-cache UV_PYTHON_INSTALL_DIR=/tmp/uv-python uv run --python 3.11 --with '.[dev,core]' pytest tests/unit/test_daily_steps.py tests/unit/test_background_steps.py tests/unit/test_config_parser.py -q`
- `UV_CACHE_DIR=/tmp/uv-cache UV_PYTHON_INSTALL_DIR=/tmp/uv-python PRE_COMMIT_HOME=/tmp/pre-commit-cache uv run --with pre-commit pre-commit run --files reme/steps/file_io/_daily_index.py reme/steps/file_io/__init__.py reme/steps/file_io/daily_write.py reme/steps/evolve/auto_memory.py reme/steps/evolve/auto_memory.yaml reme/config/default.yaml tests/unit/test_daily_steps.py tests/unit/test_background_steps.py tests/unit/test_config_parser.py docs/zh/quick_start.md docs/zh/auto_memory.md`

### DIFF
--- a/docs/zh/auto_memory.md
+++ b/docs/zh/auto_memory.md
@@ -53,6 +53,19 @@ daily/2026-06-20/session-a.md
 这样不同对话不会混在一起。一次需求讨论、一次问题排查、一次文档修改，都可以拥有自己的记忆卡片。以后想知道这一天发生了什么，先看
 `YYYY-MM-DD.md`；想看某段对话沉淀了什么，再进入对应的 `<session_id>.md`。
 
+默认情况下，`auto_memory` 使用应用时区中的运行日期作为 daily 日期。导入历史对话、评测数据或离线 transcript 时，可以传入
+`date=YYYY-MM-DD`，让 daily note 写入对话实际发生的日期：
+
+```bash
+reme auto_memory \
+  date=2023-01-19 \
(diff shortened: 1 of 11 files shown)


jinliyl:
Thanks for the update. This looks like a useful first step.

My understanding is that this PR currently only adds a \Wed Jul  1 11:21:12 CST 2026 parameter and uses it to decide the target daily note path. For benchmark or historical conversation imports, we may need to go one step further and support time information on each message as well.

Some benchmarks can contain multiple messages from the same conversation with different timestamps, so a single daily \Wed Jul  1 11:21:12 CST 2026 may not be enough to fully preserve the original temporal information. We can keep this PR focused on the daily date support, but I think we should also consider adding message-level time/timestamp support in a follow-up.

PR #312, opened 2026-07-01
feat(cli): route bare commands to the running server's real config

ployts:
call_server now resolves the client backend/transport/host/port from the live `reme start` process (replaying its start args through resolve_app_config) so a bare `reme <action>` reaches the server however it was actually launched, falling back to local config when none runs. Explicit backend=/transport=/host=/port= still win.

Also fix as_embedding to pass `dimensions` explicitly for agentscope
>=2.0.2, and add Claude Code auto-memory/auto-dream demos to the READMEs.


jinliyl (approved):
LGTM

PR #313, opened 2026-07-01
fix(auto-memory): preserve temporal grounding in notes

RerankerGuo:
## Summary
- add explicit temporal grounding rules to the auto-memory prompt
- require created notes to store conversation_date metadata
- document how conversation date, event date, and relative time should be preserved for historical imports

Refs #302

## Verification
- .......                                                                  [100%]
7 passed in 0.91s
check yaml...............................................................Passed

### DIFF
--- a/docs/zh/auto_memory.md
+++ b/docs/zh/auto_memory.md
@@ -95,9 +95,9 @@ reme auto_memory \
   messages='[{"role":"user","content":"Jon lost his job today."}]'
 ```
 
-生成记忆时，Auto Memory 还会要求正文显式区分 `conversation date`、`event date` 和 `relative time`。例如
-`"today"` 会尽量按消息时间解析为事件日期，而不是按系统运行日期解释；不能确定的相对时间会保留原始表达并标明不确定性。
-新建 daily note 时也会写入 `conversation_date` frontmatter，便于后续检索或评测对齐。
+生成记忆时，Auto Memory 会在相关时清楚保留时间事实，避免把 `"today"`、`"yesterday"`、`"recently"` 等相对时间按系统运行日期误解。
+对有歧义的历史导入，可以在正文中使用 `conversation date`、`event date`、`relative time` 这类显式标签，但默认记忆格式不会强制每条
+note 都套用固定标签。新建 daily note 时也会写入 `conversation_date` frontmatter，作为原始对话日期的 provenance 字段；它不一定等同于事件发生日期。
 
 ## 后续流向
 
--- a/docs/zh/auto_memory.md
+++ b/docs/zh/auto_memory.md
@@ -95,6 +95,10 @@ reme auto_memory \
   messages='[{"role":"user","content":"Jon lost his job today."}]'
 ```
 
+生成记忆时，Auto Memory 还会要求正文显式区分 `conversation date`、`event date` 和 `relative time`。例如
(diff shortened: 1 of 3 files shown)


jinliyl:
Thanks for the PR. I checked the change against the current auto-memory flow and the temporal-reasoning issue discussed in #302.

I agree with the general direction: for historical conversations, the memory-writing step should not silently reinterpret relative phrases such as "today", "yesterday", or "recently" against the system runtime date. Keeping the source conversation time available to the generated note is useful, and `conversation_date` metadata can also make later retrieval/evaluation more controllable.

That said, I am a little concerned that the current prompt wording may be too benchmark-oriented for the default `auto_memory` behavior. Requiring the note body to explicitly use labels like `conversation date`, `event date`, and `relative time` is helpful for LoCoMo/LongMemEval-style evaluation, but it may make everyday generated memories more rigid than necessary.

Could we consider one of these adjustments before merging?

1. Keep the general temporal-grounding instruction in the default prompt, but make the exact labels optional, e.g. "preserve temporal facts clearly when relevant" rather than requiring labeled fields in every note.
2. Move the stronger benchmark-style prompt into an evaluation/custom config, where users can override `auto_memory_step.prompt_dict` for LongMemEval/LoCoMo runs. The code already supports this: class YAML prompts are loaded first, and step-level `prompt_dict` overrides the same prompt keys.
3. If we keep `conversation_date` metadata in the default prompt, please make sure the behavior is documented as a general note provenance field, not only as an evaluation aid.

So I think the core idea is valid, but I would prefer to avoid making the default memory format too test-set-specific. A lighter default prompt plus a stronger evaluation prompt would probably give us the best tradeoff.

RerankerGuo:
Thanks for the detailed review. I updated the PR to keep the default auto-memory behavior less benchmark-specific:

- relaxed the default prompt from requiring labeled fields in every note to preserving temporal facts clearly when relevant;
- kept explicit labels such as `conversation date` and `event date` as optional guidance for ambiguous historical imports;
- documented `conversation_date` as source-conversation provenance, not necessarily the event date;
- updated the regression test to assert the lighter default behavior.

Verification:

- `PYTHONPATH=$PWD UV_CACHE_DIR=/tmp/uv-cache UV_PYTHON_INSTALL_DIR=/tmp/uv-python uv run --python 3.11 --with '.[dev,core]' pytest tests/unit/test_evolve_utils.py -q`
- `PRE_COMMIT_HOME=/tmp/pre-commit-pr313 PYLINT_HOME=/tmp/pylint-cache UV_CACHE_DIR=/tmp/uv-cache UV_PYTHON_INSTALL_DIR=/tmp/uv-python uv run --with pre-commit pre-commit run --files reme/steps/evolve/auto_memory.yaml tests/unit/test_evolve_utils.py docs/zh/auto_memory.md`

jinliyl:
Thank you for the contribution and for the careful work on temporal grounding.

After reviewing it, we feel this implementation is still a bit too benchmark-oriented for the direction we want to take in Auto Memory. We are planning to introduce a new, more general time-handling approach for the LongMemeval dataset soon, so we would prefer to wait for that design before merging changes in this area.

Thanks again for the thoughtful PR and the tests. We appreciate the effort.

RerankerGuo:
Thanks for the update on the planned time-handling approach. Closing this PR in favor of the new design you described — once #326 and any follow-up LME/automatic-memory steps stabilize, happy to re-target any remaining gaps (prompt wording, `conversation_date` semantics, regression coverage) as smaller follow-ups that align with the new direction rather than competing with it.

PR #315, opened 2026-07-01
fix(search): boost explicit temporal matches

RerankerGuo:
## Summary
- extract explicit dates from search queries, including ISO dates and common English date forms
- apply a small temporal score boost when chunk path, metadata, or text mentions the same date
- expose temporal score contributions in search output metadata/answer lines

Refs #302

## Verification
- .....                                                                    [100%]
5 passed in 0.86s
check yaml...........................................(no files to check)Skipped

### DIFF
--- a/docs/zh/memory_search.md
+++ b/docs/zh/memory_search.md
@@ -103,6 +103,7 @@ search:
     - backend: search_step
       vector_weight: 0.7
       candidate_multiplier: 3.0
+      temporal_boost: 0.0
       expand_links: true
       max_links_per_direction: 10
 ```
@@ -137,6 +138,11 @@ fused_score = vector_weight / (60 + vector_rank)
 ```
 
 默认 `vector_weight=0.7`，所以启用 embedding 后语义召回权重更高；关键词仍能把精确词命中的 chunk 拉上来。
+
+`temporal_boost` 默认是 `0.0`，即关闭。将它设为正数后，`search_step` 会从 query 中提取显式日期（例如 `2023-05-07`、
+`19 January 2023` 或 `January 19, 2023`），并轻微提升 path、metadata 或正文中包含同一日期的 chunk。这个机制只适合作为可选的
+reranking fallback；如果问题没有显式日期，例如“Jon 什么时候失业？”，仍应由上层 agent 或策略先推断时间范围，再结合
+`start_date` / `end_date` 过滤或多次检索。
 
 ## BM25 怎么工作
 
--- a/reme/config/default.yaml
+++ b/reme/config/default.yaml
@@ -294,6 +294,7 @@ jobs:
       - backend: search_step
         vector_weight: 0.7
         candidate_multiplier: 3.0
+        temporal_boost: 0.0
         expand_links: true
         max_links_per_direction: 10
 
--- a/reme/steps/index/search.py
+++ b/reme/steps/index/search.py
@@ -2,6 +2,7 @@
 
 import asyncio
 import datetime
+import re
 
 from ..base_step import BaseStep
 from ..file_io import extract_daily_date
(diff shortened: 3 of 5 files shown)


RerankerGuo:
Superseded by the later review update: temporal_boost is now exposed as an opt-in search step option with the default set to 0.0, so the explicit-date reranker is disabled unless configured by the caller.

jinliyl:
Thanks for the update. I checked this together with #317 and the original temporal-reasoning discussion in #302.

I understand the motivation: if the user query contains an explicit date, boosting chunks whose path/metadata/text mentions the same date can help prevent temporally compatible evidence from being ranked below a more generic semantic match. The implementation is also nicely scoped and tunable through `temporal_boost`.

My concern is that this may be a bit too heuristic as a default retrieval behavior. It only helps when the query already contains an explicit date, such as `2023-05-07` or `19 January 2023`. Many of the problematic questions are phrased without an explicit date, for example "When did Jon lose his job?" In those cases the boost does not add much, because the model first needs to discover the relevant temporal constraint rather than match one already present in the query.

Given #317 adds deterministic `start_date` / `end_date` filtering support at the search/file-store layer, I think a more robust direction may be agentic retrieval:

- have an upstream agent/step infer whether the question needs a time range or temporal decomposition;
- call search with `start_date` / `end_date` when the range is known or can be inferred;
- perform multiple searches when the question needs multi-hop temporal evidence;
- keep deterministic reranking as a small fallback rather than the main temporal-reasoning mechanism.

So I am not opposed to this PR, but I would prefer not to rely on the explicit-date boost as the primary fix for #302. If we keep it, I suggest positioning it as an optional/tunable reranking fallback, with the default kept conservative, and documenting that it only applies to explicit dates in the query. Otherwise, moving this logic into an evaluation/search strategy layer may be cleaner than baking it into the default search step.

RerankerGuo:
Thanks for the review. I updated this PR to position the explicit-date reranker as an opt-in fallback rather than default temporal-reasoning behavior:

- changed the default `temporal_boost` from `0.05` to `0.0`;
- kept the existing opt-in behavior for callers that explicitly set a positive boost;
- added a regression test that default search does not apply the heuristic boost;
- documented that this only matches explicit dates present in the query, and that questions without explicit dates should rely on upstream temporal decomposition / `start_date` / `end_date` filtering or multi-search strategy.

Verification:

- `PYTHONPATH=$PWD UV_CACHE_DIR=/tmp/uv-cache UV_PYTHON_INSTALL_DIR=/tmp/uv-python uv run --python 3.11 --with '.[dev,core]' pytest tests/unit/test_search_step.py tests/unit/test_config_parser.py::test_default_config_exposes_search_temporal_boost -q`
- `PRE_COMMIT_HOME=/tmp/pre-commit-pr315 PYLINT_HOME=/tmp/pylint-cache UV_CACHE_DIR=/tmp/uv-cache UV_PYTHON_INSTALL_DIR=/tmp/uv-python uv run --with pre-commit pre-commit run --files reme/steps/index/search.py reme/config/default.yaml tests/unit/test_search_step.py tests/unit/test_config_parser.py docs/zh/memory_search.md`

jinliyl:
Thank you for the contribution and for the clear tests and documentation.

After looking at the implementation, I think this approach is still a bit too benchmark-oriented for what we want to support long term. We are planning to introduce a new temporal handling strategy for the LongMemEval dataset, which should cover this use case in a more general way.

Please stay tuned, and thanks again for working on this.

RerankerGuo:
Thanks — understood. Closing this PR in favor of the new LongMemEval time-handling strategy you mentioned. The opt-in `temporal_boost` escape hatch on the search step stays around for now (not wired anywhere by default), so if it ends up useful as a small fallback later it will still be cheap to revisit.

PR #322, opened 2026-07-06
fix(file_io): fix risk of out-workspace paths

imrewce:
reme/steps/file_io/_path.py → resolve_path()
Use "whether the final resolved path falls within the workspace" as the sole security boundary.

workspace = workspace_path.resolve()

if p.is_absolute():
    logger.info("absolute path detected, recommending relative paths")
    target = p.resolve()
else:
    # 相对路径：逐段校验文件名组件（拦截 ..、非法字符等）
    for part in p.parts:
        err = validate_filename_component(part, kind="path component")
        if err:
            return None, err
    target = (workspace / p).resolve()

# 统一边界检查（绝对/相对路径共用）
if not is_relative_to(target, workspace):
    return None, "`path` must stay inside the workspace"

return target, None


jinliyl (approved):
LGTM

PR #323, opened 2026-07-06
fix(as_embedding): support both agentscope 2.0.2 and 2.0.3

ployts:
2.0.3 promoted `dimensions` to a required first-class constructor argument while keeping a backfill from `parameters.dimensions`; 2.0.2 has no such argument and reads `dimensions` from `Parameters`. Keep `dimensions` in `Parameters` for both versions and, when the model constructor accepts `dimensions`, pass `dimensions=None` so 2.0.3's backfill promotes it out of `parameters`.

### DIFF
--- a/reme/components/as_embedding/__init__.py
+++ b/reme/components/as_embedding/__init__.py
@@ -1,5 +1,6 @@
 """AgentScope embedding model wrappers."""
 
+import inspect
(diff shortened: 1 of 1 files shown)


jinliyl (approved):
LGTM

PR #325, opened 2026-07-06
feat(chunker): better json chunker and json chunker

xyf2020:
支持json和jsonl的chunker


jinliyl:
Thanks for the contribution. I checked the changes locally and the new tests pass, and `pre-commit run --all-files` also passes.

I found two issues worth addressing before merge:

1. In `JsonFileChunker`, line-range mapping is unreliable when the JSON contains repeated key names in different branches. `_build_line_index()` searches from the beginning of the file and records the first line containing a terminal key, so chunks with common keys such as `name` or `value` can get incorrect `start_line` / `end_line`. This affects result positioning, and also affects chunk IDs because the hash includes the line range.

2. When `convert_lists=False`, a valid top-level JSON array or scalar produces zero chunks. For example, `[ {"a": 1} ]` is dropped entirely because `_json_split()` only writes non-dict data when `current_path` is set. This can cause valid JSON files to be skipped from indexing if this public option is configured.

Could you please add coverage for these cases and adjust the implementation accordingly? Thanks again.

jinliyl (approved):
LGTM

PR #330, opened 2026-07-08
fix(embedding): reject mismatched embedding dimensions

jinliyl:
## Summary

This PR enforces strict embedding dimension validation across embedding generation, persistence, and vector search paths. Embeddings whose vector length does not match the active embedding model dimension are now rejected instead of being padded, truncated, reused, indexed, or persisted.

## Embedding Store Validation

- Add shared dimension validation in `BaseEmbeddingStore` so node embedding assignment only accepts vectors that match the configured model dimension.
- Replace local embedding normalization behavior with validation-only behavior in `LocalEmbeddingStore`.
- Stop padding short embeddings and stop truncating long embeddings, preventing silently corrupted vectors from entering the cache or downstream indexes.
- Fail the embedding health check when the provider returns a vector whose dimension differs from the configured dimension.

## File Store Persistence And Upsert Behavior

- Detect persisted chunk embeddings with stale dimensions during load and clear them before backfilling.
- Re-embed loaded chunks whose stored vectors were produced by an older or incompatible embedding model.
- Avoid reusing cached embeddings during upsert when the cached vector dimension no longer matches the active embedding store.
- Drop incoming or generated chunk embeddings with invalid dimensions instead of persisting them.
- Disable embedding-backed search paths when a query embedding returns an unexpected dimension, so search fails safely instead of comparing incompatible vectors.

## FAISS Index Consistency

- Rebuild FAISS indexes using only chunks whose embeddings match the active embedding model dimension.
- Validate FAISS sidecar live IDs against the set of dimension-compatible chunks.
- Skip wrong-dimensional chunks during incremental index sync after upsert.
- Return an empty result and disable embedding search when query embeddings are incompatible with the FAISS index dimension.
- Simplify the empty-index guard in FAISS vector search without changing behavior.

## Tests

- Add regression coverage for rejecting wrong-dimensional local embedding results.
- Verify base node embedding assignment does not attach invalid vectors.
- Verify health checks fail when providers return wrong-dimensional vectors.
- Verify stale persisted embeddings are invalidated and re-embedded on load.
- Verify cached stale embeddings are not reused during upsert.
- Verify custom embedding stores that return invalid vectors do not persist them.
- Verify FAISS rebuild skips chunks with mismatched dimensions.
- Confirm backfill is skipped and embeddings are disabled when the embedding store health check fails.

## Validation

- Unit Tests - py3.11: passed
- Unit Tests - py3.12: passed
- Unit Tests - py3.13: passed
- PR title check: passed
- Ubuntu workflow run: passed

### DIFF
--- a/reme/components/embedding_store/base_embedding_store.py
+++ b/reme/components/embedding_store/base_embedding_store.py
@@ -49,6 +49,7 @@ class BaseEmbeddingStore(BaseComponent):
         if embedding is None:
             return False
         dimensions = getattr(self, "dimensions", None)
+        # Base stores may not expose dimensions; only enforce the check when they do.
         if dimensions is None:
             return True
         return len(embedding) == dimensions
--- a/reme/components/embedding_store/base_embedding_store.py
+++ b/reme/components/embedding_store/base_embedding_store.py
@@ -44,11 +44,20 @@ class BaseEmbeddingStore(BaseComponent):
     async def get_embeddings(self, input_text: list[str], **kwargs) -> list[np.ndarray | None]:
         """Get embeddings for texts."""
 
+    def _embedding_dim_matches(self, embedding: np.ndarray | None) -> bool:
+        """Return whether an embedding matches the configured model dimension."""
+        if embedding is None:
+            return False
+        dimensions = getattr(self, "dimensions", None)
+        if dimensions is None:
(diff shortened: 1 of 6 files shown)


ployts (approved):
LGTM

PR #336, opened 2026-07-09
fix(cli): avoid starting unrelated components for one-shot jobs

RerankerGuo:
## Summary

- add an `Application.start_job()` path that starts only the selected job and the components referenced by that job's configured steps
- switch the CLI one-shot service to use the narrower job startup path
- add regression coverage for both no-dependency jobs and jobs that do need a referenced component

## Why

The new Windows smoke check in #335 runs `reme start service.backend=cli job=version` without embedding credentials. That command only needs `version_step`, but the previous CLI path called `app.start()` and initialized every configured component, including the default OpenAI embedding wrapper, which fails before the version job can run.

This keeps full `app.start()` behavior unchanged for long-running services while making one-shot CLI jobs avoid unrelated credential requirements.

## Verification

- `UV_CACHE_DIR=/tmp/uv-cache UV_PYTHON_INSTALL_DIR=/tmp/uv-python PYTHONPATH=$PWD uv run --python 3.11 --with '.[dev,core]' pytest tests/unit/test_job.py tests/unit/test_service.py tests/unit/test_config_parser.py -q`
- `UV_CACHE_DIR=/tmp/uv-cache UV_PYTHON_INSTALL_DIR=/tmp/uv-python uv run --python 3.11 --with '.[dev,core]' reme start service.backend=cli job=version`
- `UV_CACHE_DIR=/tmp/uv-cache UV_PYTHON_INSTALL_DIR=/tmp/uv-python PRE_COMMIT_HOME=/tmp/pre-commit-version-smoke uv run --with pre-commit pre-commit run --files reme/application.py reme/components/service/cli_service.py tests/unit/test_job.py`

### DIFF
--- a/reme/application.py
+++ b/reme/application.py
@@ -3,6 +3,7 @@
 import asyncio
 import heapq
 from concurrent.futures import ThreadPoolExecutor
+from types import CodeType
 from pathlib import Path
 from typing import AsyncGenerator, TypeVar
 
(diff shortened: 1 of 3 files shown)


jinliyl:
Thank you for investigating the CLI credential issue and contributing this change. However, I will not merge this PR.

Component startup in ReMe is configuration-driven: if a component is not needed, it should be disabled in the config. PR #341 addresses the original issue more directly by disabling embeddings in the default configuration while documenting how users can enable them when needed.

I am also concerned that the dependency inference here is incomplete. Scanning `code.co_names` does not reliably determine whether every step actually uses a component, and it does not fully cover paths such as dispatched steps or nested job calls. Introducing this inference into the application lifecycle could therefore cause subtle behavior changes.

For these reasons, we will keep the explicit configuration-based lifecycle and use #341 as the solution. Thank you again for your time and effort.

PR #337, opened 2026-07-11
fix(embedding): make input truncation CJK-aware

jinliyl:
## Summary

This PR makes embedding input truncation aware of CJK and other full-width characters, which typically consume more embedding tokens than ASCII text.

## Motivation

The existing local embedding store truncated every input by its raw Python character count. This treated ASCII and full-width text as equivalent even though CJK and other full-width characters commonly require a larger token budget.

The truncation helper also lived only on LocalEmbeddingStore, preventing other embedding store implementations from sharing the same input-length policy.

## Changes

- Move the truncation helper from LocalEmbeddingStore to BaseEmbeddingStore.
- Preserve the historical character limit and fast path for ASCII input.
- Assign a 1.5-unit cost to Unicode Wide and Fullwidth characters.
- Reserve a 5% safety margin when estimating non-ASCII input length.
- Use integer half-units to avoid floating-point boundary errors.
- Clamp negative maximum input lengths to an empty budget.
- Add regression coverage for ASCII, CJK, narrow non-ASCII, and negative-limit behavior.

## Behavior

Given a configured maximum input length of N:

- ASCII text is truncated to at most N characters, matching the previous behavior.
- Non-ASCII input uses 95% of the configured budget as an estimation safety margin.
- Narrow characters consume one budget unit.
- Wide and full-width characters consume 1.5 budget units.
- Truncation stops before the first character that would exceed the available budget.

## Testing

- uv run pytest tests/unit/test_local_embedding_store.py -q
  - Result: 4 passed
- Repository pre-commit hooks passed.
- Repository pre-push hooks passed, including credential leak detection.
- git diff --check passed.

### DIFF
--- a/reme/components/embedding_store/base_embedding_store.py
+++ b/reme/components/embedding_store/base_embedding_store.py
@@ -35,16 +35,18 @@ class BaseEmbeddingStore(BaseComponent):
     def _truncate(self, text: str) -> str:
         """Truncate text using a CJK-aware character budget.
 
-        Latin characters keep their historical cost of one unit. CJK and
-        other full-width characters cost 1.5 units because they commonly
-        consume more embedding tokens. Integer half-units avoid floating
-        point boundary errors.
+        ASCII text keeps its historical character limit. For non-ASCII text,
+        narrow characters cost one unit while CJK and other full-width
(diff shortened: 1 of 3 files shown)


ployts (approved):
LGTM

PR #338, opened 2026-07-13
fix(search): honor min_score in plain search steps

RerankerGuo:
## Summary

- apply `min_score` to the standalone vector and BM25 search steps
- filter low-scoring chunks before tool-context deduplication and result truncation
- add regression coverage for both search backends

## Root cause

`VectorSearchStep` and `Bm25SearchStep` document score filtering, but their execution paths never read `RuntimeContext.min_score`. As a result, low-confidence hits were returned even when callers supplied a threshold.

## Impact

Custom jobs and agent tools that use the plain vector or BM25 steps now get the same threshold behavior already provided by the hybrid `SearchStep`. The default threshold remains `0.0`, so existing calls without `min_score` are unchanged.

## Verification

```text
PYTHONPATH=$PWD uv run --no-sync --with '.[dev,core]' pytest tests/unit/test_search_step.py -q
# 14 passed

pre-commit run --files reme/steps/index/vector_search.py reme/steps/index/bm25_search.py tests/unit/test_search_step.py
# all hooks passed
```

### DIFF
--- a/reme/steps/index/bm25_search.py
+++ b/reme/steps/index/bm25_search.py
@@ -47,6 +47,7 @@ class Bm25SearchStep(BaseStep):
         assert self.context is not None
         query: str = (self.context.get("query", "") or "").strip()
         limit: int = int(self.context.get("limit") or 5)
+        min_score: float = float(self.context.get("min_score") or 0.0)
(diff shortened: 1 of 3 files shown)


jinliyl (approved):
LGTM

PR #341, opened 2026-07-13
chore(config): disable embeddings by default and update documentation

jinliyl:
- Set default version to 0.4.1.0
- Comment out embedding configuration in default.yaml
- Update README and README_ZH to clarify embedding components are disabled by default
- Add note explaining how to enable embedding-based semantic retrieval
- Adjust table formatting and descriptions in documentation
- Modify search command description to reflect vector search availability when enabled


ployts (approved):
LGTM

PR #361, opened 2026-07-15
refactor(embedding): defer provider construction until first remote call

jinliyl:
- Changed dimensions property to avoid forcing provider construction
- Added _ensure_model method to construct provider on demand
- Modified __call__ to ensure model exists before use
- Updated _start to defer provider initialization
- Removed eager health check during startup
- Added compact embedding serialization with base64 encoding
- Implemented batch processing for vector search with heap-based ranking
- Added document_ids property to keyword index interface
- Updated chunk persistence to handle legacy JSON embeddings
- Optimized memory usage by avoiding materialization of metadata in document_ids

### DIFF
--- a/reme/components/as_embedding/__init__.py
+++ b/reme/components/as_embedding/__init__.py
@@ -30,16 +30,26 @@ class BaseAsEmbedding(BaseComponent):
 
     @property
     def dimensions(self) -> int:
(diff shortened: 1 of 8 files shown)


ployts (approved):
LGTM

PR #365, opened 2026-07-16
feat(plugins): add Hermes Agent memory provider

### LINKED ISSUE #348: feat(plugin): add ReMe integration for Hermes Agent
**Background**

Hermes Agent supports external memory providers through its plugin system. ReMe should integrate as a native memory provider so Hermes users can retain cross-session knowledge in user-owned, file-native storage while keeping indexes and derived state rebuildable.

**Changes**
- Add an installable Hermes Agent memory provider plugin under plugins/hermes_agent
- Implement the Hermes memory provider lifecycle, including sync_turn, prefetch, shutdown, and setup integration
- Map prefetch to ReMe retrieval and sync_turn to automatic durable memory recording
- Preserve profile and conversation boundaries when routing memory operations
- Add endpoint configuration, health checks, and explicit behavior when the ReMe service is unavailable
- Add installation and setup documentation with focused lifecycle and retrieval tests

fancyboi999:
## Summary

- add a standalone Hermes Agent memory-provider plugin backed by ReMe's HTTP actions
- recall relevant context through `search` before each Hermes model call
- record completed primary-agent turns through `auto_memory` with profile- and conversation-scoped session IDs
- add profile-local setup, strict endpoint health validation, retry/degraded behavior, documentation, and contract tests
- bound inline recall separately from slow writes and drain a serial daemon writer during normal or process-exit shutdown

## Design notes

ReMe search currently operates over one workspace and does not accept a profile namespace. The plugin therefore stores its endpoint in the active `$HERMES_HOME` and documents one ReMe workspace/endpoint per Hermes profile when hard isolation is required. Conversation boundaries are preserved independently in the `auto_memory` session ID.

ReMe remains a separately managed service. If it is unavailable, Hermes startup and conversations continue, recall returns no context, and skipped or failed writes are logged explicitly. Health, retrieval, and recording use independent cooldowns so an LLM-backed write failure does not disable BM25 recall.

## Verification

- `pytest tests/unit` (`598 passed`)
- `pre-commit run --all-files`
- real ReMe HTTP service: `health_check`, `search`, and `auto_memory` created and recalled Markdown memory
- real Hermes Agent install/setup/startup: plugin installed from the repository subdirectory, setup validated the endpoint, recall affected the model response, and process-exit drain kept Hermes alive until `auto_memory` returned successfully
- two live Hermes profiles backed by separate ReMe workspaces: the seeded profile recalled the marker while the empty profile returned no memory

The external-provider integration suite was also exercised, but is not used as a merge gate here: several existing tests require provider-specific embedding configuration or model output/streaming behavior that was not satisfied by the available OpenAI-compatible endpoints.

Closes #348

### DIFF
--- a/plugins/hermes_agent/README.md
+++ b/plugins/hermes_agent/README.md
@@ -55,8 +55,10 @@ The file supports these optional settings:
 {
   "endpoint": "http://127.0.0.1:2333",
   "request_timeout": 600.0,
+  "recall_timeout": 5.0,
   "health_timeout": 2.0,
   "health_retry_seconds": 30.0,
+  "shutdown_timeout": 30.0,
   "recall_limit": 5
 }
 ```
@@ -68,11 +70,15 @@ configured. Starting a Hermes session performs a fresh endpoint health check.
 
 - `prefetch` calls ReMe `search` and returns only its recalled text. Hermes wraps
   that text in its protected memory-context block.
-- `sync_turn` sends the completed user/assistant turn to ReMe `auto_memory` with
-  a filename-safe ID derived from the Hermes profile and conversation.
+- `sync_turn` queues the completed user/assistant turn for a serial background
+  writer, which calls ReMe `auto_memory` with a filename-safe ID derived from
+  the Hermes profile and conversation.
 - Cron, flush, and subagent contexts do not write conversational memory.
 - A failed health check disables recall and recording until the retry cooldown
-  expires. Recall returns no context; a skipped or failed recording is logged as
-  a warning instead of blocking the Hermes conversation loop.
-- `shutdown` releases provider state; ReMe remains an independently managed
-  service.
+  expires. Retrieval and recording failures use independent cooldowns, so one
+  action cannot disable the other while the ReMe service remains healthy.
+- Recall uses its own short timeout so a slow ReMe search cannot stall the
+  Hermes model call for the longer automatic-memory timeout.
+- `shutdown` gives queued writes a bounded drain interval; ReMe remains an
+  independently managed service. An idempotent process-exit hook uses the same
+  drain path when a Hermes surface does not call provider shutdown directly.
--- a/plugins/hermes_agent/README.md
+++ b/plugins/hermes_agent/README.md
@@ -1,0 +1,78 @@
+# ReMe memory provider for Hermes Agent
+
+This plugin connects Hermes Agent to a running ReMe HTTP service. It recalls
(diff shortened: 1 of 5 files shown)


ployts (changes requested):
Thanks for adding the Hermes Agent memory provider. The overall integration and the focused tests look good. I found one reliability issue worth addressing before merge:

[P1] Keep the background writer alive after an unexpected exception

In plugins/hermes_agent/__init__.py:339-351, _write_loop() calls _record_payload() without catching generic exceptions. ReMeHttpClient.call() handles the common HTTP/URL/timeout/JSON errors, but exceptions such as http.client.IncompleteRead or BadStatusLine can still propagate.

If one payload triggers such an exception, the writer thread exits permanently while _write_thread remains non-None. _enqueue_write() then continues to enqueue payloads without restarting the worker, so later auto_memory writes can be silently lost until the provider is reinitialized.

Could you catch and log exceptions around each payload in _write_loop(), and also check self._write_thread.is_alive() in _enqueue_write() so a dead worker is recreated? Clearing the thread state when the loop exits would also make the lifecycle easier to reason about.

The focused plugin tests pass (15 passed), and I did not find other obvious issues in the normal execution paths.

fancyboi999:
Addressed the writer-lifecycle reliability issue in `4a0d20ac`.

- isolate unexpected exceptions per payload so one malformed HTTP response cannot terminate the writer;
- recreate the worker when `_write_thread` points to a dead thread;
- clear the thread reference on loop exit without letting an older worker clear a replacement;
- add regressions for `http.client.IncompleteRead` recovery and stale-worker restart.

Validation:
- focused Hermes plugin tests: `17 passed`;
- full unit suite: `600 passed`;
- `pre-commit run --all-files`: passed;
- post-fix live smoke with Hermes Agent v0.16.0 and ReMe v0.4.1.1: provider activated, `/search` returned recalled context, and process-exit drain completed `/auto_memory` with HTTP 200.

@ployts Could you please take another look when convenient?

ployts (approved):
Thanks for addressing the writer lifecycle issue. I reviewed the latest commit and confirmed that:

- unexpected exceptions are isolated per payload so the writer continues;
- a dead writer is detected and recreated;
- thread cleanup uses an identity check, so an older worker cannot clear a replacement.

Validation on the latest PR source:
- focused Hermes plugin tests: 17 passed;
- unit suite: 597 passed, 3 skipped;
- relevant pre-commit checks: all passed.

I did not find any remaining blocking issues. Approving.

PR #367, opened 2026-07-17
refactor(config): remove max_file_bytes limit from background jobs

jinliyl:
- Removed max_file_bytes configuration from index_update_loop, resource_watch_loop, digest_watch_loop, and reindex jobs
- Updated default.yaml to reflect simplified job configurations without file size limits
- Removed corresponding test case that validated the 20 MiB limit behavior
- Simplified watch directories and suffixes to basic configurations
- Cleaned up unnecessary commented configurations in the YAML file

### DIFF
--- a/reme/config/default.yaml
+++ b/reme/config/default.yaml
@@ -4,11 +4,8 @@ service:
 jobs:
   index_update_loop:
     backend: background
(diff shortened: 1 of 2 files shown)


ployts (approved):
LGTM

PR #372, opened 2026-07-17
feat(plugin): add ReMe integration for Codex

### LINKED ISSUE #350: feat(plugin): add ReMe integration for Codex
**Background**

Codex users need a first-class way to use ReMe as persistent, file-native memory across coding sessions. The existing `reme_memory` skill provides a generic CLI workflow, but a dedicated Codex plugin should package the skill and service connection for lower-friction installation while preserving user-owned ReMe workspace files as the source of truth.

**Changes**
- Add an installable Codex plugin under `plugins/codex` with plugin and marketplace metadata
- Bundle a ReMe memory skill for semantic search, graph traversal, state lookup, and source-backed reads
- Connect Codex to the shared ReMe MCP service for structured recall and memory operations
- Integrate supported Codex lifecycle hooks to record durable session knowledge without blocking session completion
- Preserve project, workspace, and conversation boundaries when routing memory operations
- Add endpoint configuration, health checks, and explicit behavior when the ReMe service is unavailable
- Add installation documentation and focused tests for plugin discovery, recall, and automatic recording

iluv7:
## Description

Add an installable Codex plugin (`plugins/codex/`) that connects Codex to ReMe for persistent, file-native long-term memory.

Closes #350.

## What Changed

- **MCP** (`.mcp.json`) — connects Codex to the ReMe MCP server, exposing 27+ memory tools
- **Recall skill** (`SKILL.md`) — teaches Codex when and how to recall from ReMe
- **Stop hook** (`hooks.json` + `auto_memory.py`) — calls `auto_memory_cc` per turn via double-fork, never blocks
- **Plugin manifest** (`plugin.json`) — identifies the plugin, wires up skill/MCP/hooks
- **Marketplace** (`marketplace.json`) — repo-root catalog so `codex plugin marketplace add` discovers the plugin
- **Docs** (`README.md`) — installation and verification instructions

The plugin follows the same architecture as the existing `plugins/claude_code/` — identical `.mcp.json` format, same `SKILL.md` content, same hook logic. Codex-specific delta: `plugin.json` `interface` block, `PLUGIN_ROOT` env var.

### 1. Marketplace registration

4bc0f5c01eafc0e535dd5434477365fa

The repo-root `marketplace.json` is discovered correctly. Codex reads the plugin entry pointing to `./plugins/codex/reme`.

### 2. Plugin installation
image

- **Method**: `codex plugin marketplace add .` → Codex session → `/plugins` → Install "ReMe Memory"
- **Result**: plugin installed at `~/.codex/plugins/cache/reme-marketplace/reme/0.1.0/`
- **Config verification**: `~/.codex/config.toml` shows `[plugins."reme@reme-marketplace"] enabled = true`

### 3. Skill loading
f5103621a773bb00d51e3b54ec3f81c4

- **Verified**: `reme-memory` skill appears in the Codex skill list

### 4. MCP connection
700cd10ba4178ad33a1088190b303f2d

- **Server registration**:
  ```bash
  $ codex mcp list
  Name  Url                        Status   Auth
  reme  http://127.0.0.1:8000/mcp  enabled  Unsupported
  ```

- **Tool availability**: 27 MCP tools confirmed via `tools/list`:
  ```
  auto_dream, auto_memory, auto_memory_cc, auto_resource, proactive,
  version, health_check, status, help, traverse, reindex, search,
  node_search, daily_list, daily_reindex, frontmatter_delete,
  frontmatter_read, frontmatter_update, stat, list, move, delete,
  read, read_image, write, daily_write, edit
  ```

- **Live calls verified**:
  - `health_check` → returns healthy component snapshot
  - `version` → returns ReMe v0.4.1.1
  - `list_mcp_resources` → correctly queries the `reme` server

### 5. Stop hook — automatic recording

- **Hook registration**:
  ```json
  // hooks/hooks.json
  { "hooks": { "Stop": [{ "hooks": [{ "type": "command", "command": "python3 \"${PLUGIN_ROOT}/hooks/auto_memory.py\"", "timeout": 30 }] }] } }
  ```

- **Hook execution log** (`logs/auto_memory_hook.log`):
5aa3fdd61716fb073fe85df7cd3dc724

The hook's `ok` status also serves as an independent verification of the MCP connection — `auto_memory.py` is itself a full MCP JSON-RPC client (initialize → initialized → tools/call), so each `ok` confirms the entire MCP protocol stack is working end-to-end.


ployts (commented):
Thanks for adding the Codex integration. I reviewed the plugin packaging and checked the Codex-specific behavior against the current hook/plugin documentation. There are a few issues to address:

1. The bundled MCP endpoint uses http://127.0.0.1:8000/mcp, but ReMe defaults to port 2333 and the README tells users to start it without a port override. A fresh install following the documentation therefore cannot connect for either recall or recording.

2. The Stop hook passes the Codex session ID to auto_memory_cc. That job is Claude Code-specific: it searches ~/.claude/projects for the transcript and parses Claude Code records. Codex already supplies transcript_path in the hook payload, but the hook discards it. For a normal Codex session this produces no messages, AutoMemoryStep returns "Skipped: no messages", and the hook still logs the JSON-RPC response as "ok". Automatic recording therefore does not currently record Codex sessions. This needs a Codex-aware transcript ingestion path, plus validation of the actual tool result.

3. The installation instructions should tell users to open /hooks and explicitly trust the plugin hook. Codex skips non-managed plugin hooks until they are reviewed and trusted, so automatic recording is otherwise disabled after installation.

4. Native Windows also needs handling. The hook command uses python3 without commandWindows. Even if Python is found, os.fork is unavailable, so the operation runs inline with a 600-second request timeout while Codex kills the hook after 30 seconds. Please add a detached Windows command path or document the supported platforms.

The static checks pass (diff check, JSON parsing, and Python syntax), and CI is green. However, the PR currently adds no focused tests for plugin discovery or automatic recording, so these runtime paths are not covered.

iluv7:
@ployts 
Thank you for the thorough review. These issues are entirely my oversight — I didn't properly account for the differences between Codex and Claude Code's hook payload and transcript handling. Several things were not correctly adapted. My apologies.

Here's a summary of the changes:

1. Port mismatch
.mcp.json hardcoded port 8000, but ReMe defaults to 2333. Fixed to 2333, now consistent with the README, SKILL.md, and the _server_url() fallback.

2. Codex transcript ingestion (core)
The hook was previously calling auto_memory_cc, which searches ~/.claude/projects/ for transcripts and parses user/assistant format entries. Codex stores transcripts under $CODEX_HOME/sessions/ and uses a different format (response_item / developer / input_text). Added a new AutoMemoryCodexStep:
- Reads directly from transcript_path provided by the hook payload, instead of searching ~/.claude/projects/
- Parses the Codex-native format
- Falls back to searching $CODEX_HOME/sessions/ when the given path doesn't exist (defensive handling for known worktree path bugs)
- Deduplicates by uuid so only new turns since the last stop are processed

The hook was also updated: reads transcript_path, calls auto_memory_codex instead of auto_memory_cc, and validates the result by checking n_messages metadata rather than just looking for an error field.

3. Windows support
Added commandWindows to hooks.json and _spawn_detached() to auto_memory.py using subprocess.Popen with DETACHED_PROCESS for fire-and-forget background recording.

4. README
Added hook trust step to installation instructions and a supported platforms section.

5. Tests
Added 39 unit tests across 3 files covering plugin structure validation, step logic (message rendering, content parsing, transcript path resolution, dedup), and hook behavior (result classification, payload parsing, Windows detach).

ployts (commented):
Thanks for the update. I rechecked commit `61e5871`. The port, hook-trust instructions, dedicated Codex step, and focused tests are good improvements, but there are still several issues to address before merge:

1. **P1 — the parser does not match the persisted Codex rollout schema.** A rollout row is `{"type":"response_item","payload":{...}}`, and the message is inside `payload` with `type: "message"`, role `user`/`assistant`, and content items such as `{"type":"input_text","text":"..."}` or `{"type":"output_text","text":"..."}`. There is no top-level `uuid` or `message`. As a result, [the uuid filter](https://github.com/agentscope-ai/ReMe/blob/61e58717d7f6b62e33c4a62338d516a8f81a6669/reme/steps/evolve/auto_memory_codex.py#L169-L173) drops all real rows, and [the renderer](https://github.com/agentscope-ai/ReMe/blob/61e58717d7f6b62e33c4a62338d516a8f81a6669/reme/steps/evolve/auto_memory_codex.py#L183-L217) reads the wrong nesting. The current tests use an invented top-level `uuid/message` shape, so please replace/add a fixture matching the official [RolloutItem envelope](https://github.com/openai/codex/blob/bf3c1972b7d045c0a3a48dff91f381070f8f69e1/codex-rs/protocol/src/protocol.rs#L3197-L3202) and [ResponseItem schema](https://github.com/openai/codex/blob/bf3c1972b7d045c0a3a48dff91f381070f8f69e1/codex-rs/protocol/src/models.rs#L711-L727).

2. **P1 — the Windows detached child loses its payload.** In [this block](https://github.com/agentscope-ai/ReMe/blob/61e58717d7f6b62e33c4a62338d516a8f81a6669/plugins/codex/reme/hooks/auto_memory.py#L242-L249), `sys.stdin` is assigned to `fh`, but the `with` block closes `fh` before `main()` reads it. `main()` therefore falls back to an empty payload and returns without recording. Please deserialize the file first, or invoke the recording path while the file is still open.

3. **P1 — constrain transcript reads to Codex session storage.** `auto_memory_codex` is exposed as a public MCP job, and [any existing absolute JSONL path is accepted](https://github.com/agentscope-ai/ReMe/blob/61e58717d7f6b62e33c4a62338d516a8f81a6669/reme/steps/evolve/auto_memory_codex.py#L111-L115). An MCP caller can therefore make ReMe read and send arbitrary local JSONL content to the configured model. Please resolve the path and verify that it is under `$CODEX_HOME/sessions` before opening it, or accept only an opaque session id and resolve it server-side.

4. **P2 — skipped-result classification still cannot work through the real MCP service.** [The hook checks `result.metadata.n_messages`](https://github.com/agentscope-ai/ReMe/blob/61e58717d7f6b62e33c4a62338d516a8f81a6669/plugins/codex/reme/hooks/auto_memory.py#L47-L60), but ReMe MCP returns only `response.answer` and discards `Response.metadata`. A real `Skipped: no messages` response is still logged as `ok`.

5. **P2 — the fallback filename pattern is incorrect.** Codex session files are named `rollout-<timestamp>-<session_id>.jsonl`, while [the fallback searches only for an exact `<session_id>.jsonl`](https://github.com/agentscope-ai/ReMe/blob/61e58717d7f6b62e33c4a62338d516a8f81a6669/reme/steps/evolve/auto_memory_codex.py#L124-L129).

Please also update the README sections that still describe `auto_memory_cc` and passing only a session id, and move `.agents/plugins/marketplace.json` to `plugins/codex/marketplace.json` so the Codex marketplace metadata stays under the Codex plugin subtree.

Once these are fixed with realistic rollout and Windows payload tests, I can re-review.

iluv7:
@ployts 
Thank you for the thorough re-review. All six issues should now be addressed:

P1 — Parser
Fixed to match the real Codex RolloutItem schema: reads payload nesting, uses payload.id for dedup, supports input_text/output_text blocks with the text field. Tests updated with realistic fixtures.

P1 — Windows payload
The detached child now reads the payload file into memory before the with block closes, so main() receives the full payload instead of an empty one.

P1 — Path validation
_resolve_transcript_path rejects paths outside $CODEX_HOME/sessions to prevent MCP callers from reading arbitrary files.

P2 — Result classification
_result_status now checks the answer text for "Skipped" prefix, since ReMe MCP drops Response.metadata and only returns answer.

P2 — Fallback filename
Fallback search uses rollout-*-<session_id>.jsonl to match real Codex file naming.

Other
- marketplace.json moved to plugins/codex/
- README updated to reference auto_memory_codex and transcript_path

iluv7:
Added MCP round-trip tests with a mock JSON-RPC server.

iluv7:
First, apologies for the oversight. When scaffolding the Codex plugin, I reused `AutoMemoryCCStep` as a reference since both plugins share the same architecture (MCP Stop hook → server-side step → `AutoMemoryStep`). I adjusted the nesting to read `payload.role` / `payload.content`, but missed that the content block types and tool-call representation are fundamentally different between the two formats.

### Codex vs Claude Code schema differences

I checked the Codex CLI transcript schema against [toolpath-codex types.rs](https://docs.rs/toolpath-codex/latest/src/toolpath_codex/types.rs.html).

| Aspect | Claude Code | Codex |
|---|---|---|
| Top-level | `{type: "user"/"assistant", message: {role, content}, uuid}` | `{type: "response_item", payload: {type, ...}}` |
| Content blocks | `text`, `tool_use`, `tool_result` | Only `input_text`, `output_text` |
| Tool calls | Inline blocks inside assistant message content | Separate `response_item` entries: `function_call`, `function_call_output` |
| Injection tags | `<bash-input>`, `<system-reminder>`, … | `<user_instructions>`, `<environment_context>`, … |

### Issues found

1. **Tool calls are completely lost.** Only `payload.type == "message"` was handled. The four tool-related payload types (`function_call`, `function_call_output`, `custom_tool_call`, `custom_tool_call_output`) were silently dropped — shell commands, MCP tools, and file operations were never recorded into memory.

2. **Dedup key only used `payload.id`.** `function_call_output` has no `id` field, only `call_id`, so these entries were filtered out during dedup even before the rendering step.

3. **Content block types `"text"`, `"tool_use"`, `"tool_result"` don't exist in Codex `ContentPart`.** These branches were copied from CC and are unreachable on real Codex data.

4. **Injection filtering uses CC tag names** (`<bash-input>`, `<system-reminder>`, etc.) that never appear in Codex transcripts, so injected system content passes through unfiltered.

### Changes (2 files, ~300 lines rewritten)

**`reme/steps/evolve/auto_memory_codex.py`**

Removed:
- `_INJECTED_TAGS` constant and `_is_injected_only()` — CC-specific, doesn't apply to Codex
- `"text"`, `"tool_use"`, `"tool_result"` block types — don't exist in Codex

Added:
- `_render_function_call_payload()` — renders tool calls (shell, MCP, file ops) as `[tool {name}({args})]`
- `_render_function_call_output_payload()` — renders tool output as `[tool_result {excerpt}]`
- `_render_custom_tool_call_payload()` / `_render_custom_tool_call_output_payload()`
- Payload `type` dispatch in `_render_codex_payload()` — routes to the correct renderer per type
- Module docstring documenting the full Codex rollout schema

Fixed:
- `_save_codex_session()` dedup key: `payload.id` → `payload.id or payload.call_id`

**`tests/unit/test_auto_memory_codex.py`**

- Helpers updated to match the real `response_item` envelope
- Added: `function_call` / `function_call_output` / `custom_tool_call` rendering, `reasoning` skip, `call_id`-based dedup
- Added `test_realistic_multi_turn_session` — 11-entry transcript (session_meta, turn_context, 4 messages, 2 function_calls, 2 function_call_outputs, 1 reasoning) covering the full parse chain
- Trimmed trivial assertions; 19 tests total

### Test results

```
48 passed (19 auto_memory_codex + 11 codex_plugin + 11 hook_auto_memory + 7 codex_hook_mcp)
```

iluv7:
Ran the full write + recall pipeline end to end, everything passed.

### Environment

- macOS 15.3.1 (Apple Silicon), Codex CLI, ReMe v0.4.1.1 (local dev)
- LLM: DeepSeek v4-pro (`api.deepseek.com`). Embedding: not configured (DeepSeek does not provide an embedding API).
- ReMe MCP server: `http://127.0.0.1:2333/mcp`

### Write session

1. User enters in a Codex session: **"我喜欢用 Go 1.22"**

2. Agent loads the `reme-memory` skill, confirms the memory system is online, then responds: **"记住了: 后续涉及 Go 版本时我会默认按 Go 1.22 来处理。"**

3. When the session ends, the Stop hook fires and calls ReMe's `auto_memory_codex` tool over MCP:

```bash
# Check hook log
cat ~/.codex/plugins/cache/reme-marketplace/reme/0.1.0/logs/auto_memory_hook.log
# 2026-07-20 23:31:11 session=019f8026-b643-7370-ae96-676fefd7b225 ok
```

4. Parser & LLM extraction on the server side — zero errors:

```bash
# Check server-side parse + LLM extraction log
grep -E "resolved Codex|auto_memory|daily.*created|Error" /tmp/reme-server.log
```

```
[AutoMemoryCodexStep] resolved Codex session
  transcript=20 new_entries=8 messages=4     ← JSONL parse + dedup + render

[AutoMemoryCodexStep] note lookup ... created=True  ← daily note created
[AutoMemoryCodexStep] agent start ... agent done has_result=True  ← LLM extraction
[AutoMemoryCodexStep] refresh index done ... go-default-version.md  ← indexed
[AutoMemoryCodexStep] done ... modified=True  ← written
```
db04a1013909552a726971c0601e9fdf


### Recall session

1. In a new session, user asks: **"我喜欢用什么编程语言 什么版本"**

2. Agent follows the skill instructions step by step:
   - `reme.version` → `0.4.1.1`
   - `reme.health_check` → healthy
   - `reme.search({"query":"user prefers programming language version","limit":5})` → hit

3. Search results (keyword-only since DeepSeek has no embedding API, `vector_hits=0`):

```bash
# Check search & recall log
grep -E "SearchStep|DailyListStep|search.py|daily_list" /tmp/reme-server.log
```

```
[DailyListStep] date=2026-07-20 notes=0    ← before write
[DailyListStep] date=2026-07-20 notes=1    ← after write
[SearchStep] query='user prefers programming language version' keyword_hits=1
[SearchStep] query='user preference programming language default version' keyword_hits=1
```

   Search response:
   ```
   daily/2026-07-20.md:5-8 [score=2.5415]
   [[daily/2026-07-20/go-default-version.md]]
   name: go-default-version
   description: User preference: Go 1.22 is the default Go version to assume for
   all Go-related tasks (build, language features, standard library, etc.).
   ```

4. Agent double-checks then responds: **"你记录里的明确默认偏好是 Go，版本是 1.22。对应记忆在 ./daily/2026-07-20/go-default-version.md。"**
998623719ef81e563864e34ea4e198b7


### Result

| Step | Status | Evidence |
|---|---|---|
| Stop hook fires & MCP connects | ✅ | `auto_memory_hook.log: ok` |
| JSONL parse, dedup & render | ✅ | `transcript=20 → messages=4` via `_codex_entries_to_messages` |
| LLM fact extraction | ✅ | `agent done has_result=True` → `go-default-version.md` |
| Memory indexed | ✅ | `refresh index done`, `DailyListStep notes=0→1` |
| Cross-session recall | ✅ | Two `search` calls both `keyword_hits=1`, agent returns Go 1.22 |

ployts (commented):
Thanks for the parser rewrite and the end-to-end validation. I rechecked commit `96a167e`. The payload dispatch, Windows fix, path confinement, fallback filename, and skipped-result handling are improved, but three issues remain:

1. **P1 — valid Codex messages without `payload.id` are still discarded before rendering.** `ResponseItem::Message.id` is optional in the official schema, and Codex itself persists user rollout messages with `id: None` ([official example](https://github.com/openai/codex/blob/bf3c1972b7d045c0a3a48dff91f381070f8f69e1/codex-rs/rollout/src/persistence_metrics_tests.rs#L29-L38)). However, [`_save_codex_session`](https://github.com/agentscope-ai/ReMe/blob/96a167e4c0303416b612b5e668f1d87f463f581f/reme/steps/evolve/auto_memory_codex.py#L164-L190) keeps only rows with `id` or `call_id`. With an id-less user message and assistant message, both are renderable before saving but the saved increment is empty, so automatic memory receives zero messages. The new fixtures still make `id` mandatory for every message. Please use a stable cursor/order key such as the rollout line position or timestamp/content fingerprint, and add an id-less user-message test. The successful Go 1.22 E2E can still pass if the assistant echoed the preference, so it does not prove that the original user row survived dedup.

2. **P1 — the moved marketplace manifest is not discoverable by Codex.** I need to correct my earlier relocation suggestion: current Codex only recognizes marketplace manifests at [the supported relative paths](https://github.com/openai/codex/blob/bf3c1972b7d045c0a3a48dff91f381070f8f69e1/codex-rs/core-plugins/src/marketplace.rs#L20-L25), including `.agents/plugins/marketplace.json`; `plugins/codex/marketplace.json` is not one of them. Therefore the documented clean install command, `codex plugin marketplace add agentscope-ai/ReMe`, cannot discover the manifest at the current head. The E2E log points to an already installed cache entry, so it does not exercise a clean marketplace add from this layout. Please restore a supported discoverable manifest (or add a packaging step that materializes one) and test installation from a clean marketplace source.

3. **P2 — ReMe job failures are still logged as `ok`.** [`_result_status`](https://github.com/agentscope-ai/ReMe/blob/96a167e4c0303416b612b5e668f1d87f463f581f/plugins/codex/reme/hooks/auto_memory.py#L47-L69) recognizes JSON-RPC errors and `Skipped`, but ReMe MCP returns `response.answer` as normal text even when `response.success` is false. For example, `Error: could not resolve transcript path` is classified as `ok`. Please classify the returned error answer or preserve the success/error state through the MCP boundary, and cover this real response shape.

Minor documentation cleanup: the README notes still refer to `plugins/reme/.mcp.json` and `plugins/reme/logs/`, which do not match this subtree.

Validation on `96a167e`: the 48 focused tests pass and `git diff --check` passes; the id-less rollout and normal-content error cases above fail as described.

iluv7:
## P1 — id-less messages discarded: FIXED ✅

**Root cause**: `_save_codex_session` only kept entries with `payload.id` or `payload.call_id`. But Codex `ResponseItem::Message.id` is optional — Codex persists user rollout messages with `id: None`. Id-less entries were silently dropped, leaving auto memory with zero messages.

**Fix**: Added `_entry_dedup_key()` with three-tier fallback: `payload.id` → `payload.call_id` → content hash (`hash(json.dumps(entry, sort_keys=True))`). Id-less entries now get a stable fingerprint instead of being discarded.

**Tests**: 
`test_id_less_entries_not_discarded` (id-less entries are kept), 
`test_id_less_message_rendered_and_deduped` (same content deduped), `test_different_id_less_messages_not_collapsed` (different content not falsely merged).

## P1 — Marketplace manifest not discoverable: FIXED ✅

- Deleted `plugins/codex/marketplace.json` — Codex does not discover manifests at this path.
- `.agents/plugins/marketplace.json` (a supported path) already exists, `source.path: "./plugins/codex/reme"` resolves correctly relative to repo root.

**Clean install verification**:

```bash
codex plugin marketplace remove reme-marketplace
rm -rf ~/.codex/plugins/cache/reme-marketplace

codex plugin marketplace add .
# Added marketplace `reme-marketplace` from /Users/guoyuchen/workspace/Agent/ReMe.

codex plugin add reme --marketplace reme-marketplace
# Added plugin `reme` from marketplace `reme-marketplace`.
# Installed plugin root: ~/.codex/plugins/cache/reme-marketplace/reme/0.1.0

codex plugin list
# reme@reme-marketplace  installed, enabled  0.1.0  .../plugins/codex/reme
```

## P2 — ReMe errors logged as ok: FIXED ✅

`reme/components/service/mcp_service.py`, `plugins/codex/reme/hooks/auto_memory.py`.

**Root cause**: Two bugs compounded. (1) Server: `mcp_service.py` returned `response.answer` unconditionally — even when `response.success` was `False`, the MCP transport layer saw a normal function return and produced `isError: false`. (2) Hook: `_result_status` checked JSON-RPC `"error"` and `"Skipped"` text prefix but ignored `isError` in the result, so errors that did surface as `isError: true` were still logged as `ok`.

**Fix**: Server: `execute_tool` raises `ToolError(str(response.answer))` when `response.success` is `False`. FastMCP converts this to `CallToolResult(isError=true)`. Hook: `_result_status` now checks `"error" in result` → `isError` → `"Skipped"` prefix → `ok`. Extracted text parsing into `_extract_text()` shared by `_result_status` and `_error_detail`. `_mcp_call` catches `HTTPError` on `tools/call` so non-2xx error responses can still be parsed.

**Tests**: `test_fail_job_raises_tool_error` (MCPService `execute_tool` with `success=False` → `ToolError`), `test_realistic_fail_job_raises_tool_error` (real `auto_memory_codex.py:64-66` pattern), `test_error_on_mcp_is_error` (`isError: true` → error).

**Manual verification** (against a running ReMe server on 2333):
```
$ curl $URL -d '{"method":"tools/call","params":{"name":"auto_memory_codex",
  "arguments":{"transcript_path":"/nonexistent","session_id":"test-error"}}}'
// Before: {"result": {"content": [{"text": "Error: ..."}], "isError": false}}
// After:  {"result": {"content": [{"text": "Error: ..."}], "isError": true}}

$ hook log:
// Before: session=test-error ok
// After:  session=test-error error Error: could not resolve transcript path
```

## Doc paths: FIXED ✅

- `plugins/codex/README.md`: `plugins/reme/.mcp.json` → `plugins/codex/reme/.mcp.json`, 
- `plugins/reme/logs/` → `plugins/codex/reme/logs/`
- `plugins/codex/reme/skills/reme-memory/SKILL.md`: `plugins/reme/.mcp.json` → `plugins/codex/reme/.mcp.json`

iluv7:
Force push: cleaned up branch history for review

- Rebased onto latest main — dropped merge commits, now a clean linear history.
- Squashed 9 noisy commits (revert round-trip, overlapping changes) into one.

Just review ba97f5d — it covers all four fixes (id-less messages, marketplace manifest, isError handling, doc paths) across 9 files.

ployts (approved):
Reviewed ba97f5d. The previously blocking id-less message handling and marketplace discovery issues are fixed. The remaining Windows detach behavior and default MCP error-status reporting can be handled as non-blocking follow-ups. Approving.

PR #377, opened 2026-07-19
fix(mcp): preserve response metadata in MCP tool results

### LINKED ISSUE #376: bug: MCP tool responses lose structured metadata — only human-readable text is returned
## Description

When calling ReMe tools via MCP protocol (SSE/stdio/streamable-http), the structured
metadata in `Response.metadata` is silently discarded. Clients receive only the
human-readable `answer` string, making it impossible to programmatically access file
lists, search results, graph edges, etc.

## Affected Tools

| Tool          | What client receives                 | Lost metadata                                |
|---------------|--------------------------------------|----------------------------------------------|
| `list`        | `"Listed 4 file(s) under user-001/"` | `items: string[]`, `count: int`              |
| `search`      | Human-readable search text           | `results`, `link_expansion`, `counts`        |
| `daily_list`  | Human-readable note list             | `notes`, `count`, `date`                     |
| `traverse`    | Graph traversal text                 | `edges`, `count`                             |
| `node_search` | Node list text                       | `hits`, `counts`                             |
| `proactive`   | Read result summary                  | `topics`, `content`, `skipped`               |

All MCP clients (not limited to any specific client) are affected.

## Root Cause

In `reme/components/service/mcp_service.py`, the `execute_tool` closure inside
`MCPService.add_job()` returns only `response.answer`:

```python
async def execute_tool(**kwargs):
    response = await job(**kwargs)
    return response.answer  # ← metadata discarded here
```
The HTTP Service correctly returns the full Response JSON (with answer + metadata), but the MCP Service does not.

## Expected Behavior
MCP clients should be able to access structured metadata from tool responses, consistent with what the HTTP Service provides.

## Environment
ReMe version: 0.4.x
Transport: SSE (also affects stdio and streamable-http)
Python: 3.12

---

LiangshouX:
## Problem

`MCPService.add_job()` returns only `response.answer` when registering jobs as
MCP tools, discarding `response.metadata` that carries structured data.

Affected tools: `list`, `search`, `daily_list`, `traverse`, `node_search`,
`proactive` — all tools that populate `response.metadata` with structured
results.

MCP consumers (e.g. agents, external integrations) receive human-readable text
instead of structured data, making programmatic access impossible.

## Root Cause

In `mcp_service.py`, the `execute_tool` closure:

```python
async def execute_tool(**kwargs):
    response = await job(**kwargs)
    return response.answer  # metadata discarded
```
## Fix

When `response.metadata` is non-empty, return a JSON envelope:

```python
if response.metadata:
    return json.dumps(
        {"answer": response.answer, "metadata": response.metadata},
        ensure_ascii=False,
        default=str,
    )
return response.answer
```

This preserves backward compatibility: tools with empty metadata return the same
plain text as before.

## Testing

- Added `tests/unit/test_mcp_service_metadata.py` with 3 test cases:
  - Empty metadata → plain answer string (unchanged behavior)
  - Non-empty metadata → JSON envelope with answer + metadata
  - Non-serializable metadata types → graceful fallback via `default=str`
- All existing `tests/unit/test_service.py` tests pass
- `pre-commit run --all-files` passes

## Impact

- **MCP consumers**: Can now access structured data (file lists, search results)
  from tools that populate metadata
- **HTTP consumers**: Unaffected (separate code path)
- **Agents/LLM**: Receive JSON instead of plain text for metadata-rich tools;
  JSON is equally or more parseable by LLMs

Fixes #376

### DIFF
--- a/reme/components/service/mcp_service.py
+++ b/reme/components/service/mcp_service.py
@@ -139,12 +139,23 @@ class MCPService(BaseService):
     def add_job(self, job: BaseJob) -> bool:
         """Register a non-stream job as an MCP tool; StreamJobs are unsupported."""
         from fastmcp.tools import FunctionTool
+        import json as _json
 
         if isinstance(job, StreamJob):
             return False
 
         async def execute_tool(**kwargs):
             response = await job(**kwargs)
+            # When metadata carries structured data, return a JSON envelope
+            # so MCP consumers can access file lists, search results, etc.
+            # For tools whose metadata is empty the plain answer string is
+            # returned as before.
+            if response.metadata:
+                return _json.dumps(
+                    {"answer": response.answer, "metadata": response.metadata},
+                    ensure_ascii=False,
+                    default=str,
+                )
             return response.answer
 
         self.service.add_tool(
--- a/tests/unit/test_mcp_service_metadata.py
+++ b/tests/unit/test_mcp_service_metadata.py
@@ -1,0 +1,112 @@
+"""Tests for MCP service metadata preservation."""
+
+import json
+from types import SimpleNamespace
+from unittest.mock import AsyncMock, patch
+
+import pytest
+
+from reme.components.job import BaseJob
+from reme.components.service.mcp_service import MCPService
+from reme.schema import Response
+
+
+def _dummy_app():
+    """Minimal object needed by MCPService.build_service."""
+
+    async def start():
+        return None
+
+    async def close():
+        return None
+
+    return SimpleNamespace(
+        config=SimpleNamespace(app_name="test"),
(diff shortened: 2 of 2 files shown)


jinliyl:
Thank you for the contribution and for adding focused tests.

The current MCP behavior is intentional: it is a direct-to-LLM interface, so Response.answer is the tool output, while Response.metadata remains auxiliary request data for programmatic clients and diagnostics. Automatically serializing metadata into MCP output would mix those responsibilities and change what the LLM receives.

When a tool needs to expose additional information to the LLM, we prefer updating that Step to include it in answer. For that reason, we do not plan to apply this transport-level change as proposed.

PR #389, opened 2026-07-23
fix(file_io): stop ReadStep small-file path over-counting total lines by 1

### LINKED ISSUE #388: fix(file_io): ReadStep small-file path over-counts total lines by 1 for trailing-newline files
`ReadStep`'s small-file path (`reme/steps/file_io/read.py:114`) computes the file's
line count with:

```python
all_lines = content.split("\n")
total = len(all_lines)
```

For any file ending in a newline (the normal case for well-formed text), `split("\n")`
produces one extra, empty trailing element, so `total` is inflated by 1. A 3-line file
`"line1\nline2\nline3\n"` reports `total=4`.

This is inconsistent with the other two line-counting paths over the same content:
`read_file_lines_safe` (the large-file path, `reme/steps/file_io/_file_io.py:132`,
`async for line in f`) and `default_file_chunker`'s chunker
(`reme/components/file_chunker/default_file_chunker.py:129-131`, which explicitly
decrements when the last byte is a newline). Both report 3 for the same file.

Concrete symptom: `_resolve_range` rejects `start_line` past EOF, but the inflated
total makes a request for `start_line = <real_total> + 1` (one line past the actual
end of file) succeed instead, returning an empty excerpt with `success=True`.

Repro (against a fresh `LocalFileStore` + `ReadStep`, real production code, no mocks):

```python
content = "line1\nline2\nline3\n"  # real line count: 3
# ReadStep(...).execute() with start_line=99 -> "start_line 99 exceeds file length (4 lines)"
# ReadStep(...).execute() with start_line=4  -> success=True, answer='' (should be rejected)
```

I have a fix and a regression test ready and will open a PR.

AmirF194:
`ReadStep`'s small-file path computes the line count with `content.split("\n")`, whose
length includes a trailing empty element whenever the file ends in a newline
(`read.py:114`). That inflates `total` by 1 for every well-formed text file, so
`_resolve_range` accepts `start_line = <real_total> + 1`, one line past the actual end
of the file, and returns an empty excerpt with `success=True` instead of rejecting it.

The other two line-counting paths over identical content already get this right:
`read_file_lines_safe` (large-file path, iterates the file object line by line) and
`default_file_chunker` (`default_file_chunker.py:129-131`, decrements `end_line` when
the last byte is a newline). This fix mirrors the same correction in `ReadStep`: drop
the trailing empty element from `split("\n")` when the content ends in `\n`, so the
count agrees with both.

Verified in Docker (`python:3.11-slim`, editable install) against the real production
`ReadStep`/`LocalFileStore`, no mocks: with content `"line1\nline2\nline3\n"` (3 real
lines), `start_line=99` reported `"exceeds file length (4 lines)"` on main and
`"exceeds file length (3 lines)"` on this branch; `start_line=4` (one past real EOF)
returned `success=True, answer=''` on main and is correctly rejected on this branch.
`tests/unit/test_read_trailing_newline_does_not_inflate_total_lines` (new) fails on
main and passes here. Full `tests/unit` suite: 747 passed, 1 pre-existing failure
(`test_stat_indexed_file`, a MIME-type lookup gap in the slim Docker image, reproduces
identically on unpatched main and is unrelated to this change). `pre-commit run
--files` clean on both changed files.

Not verified: behavior on Windows/other filesystems with different newline
conventions, since the fix operates on the already-decoded string content.

Fixes #388

### DIFF
--- a/reme/steps/file_io/read.py
+++ b/reme/steps/file_io/read.py
@@ -112,6 +112,12 @@ class ReadStep(BaseStep):
                 return None
 
             all_lines = content.split("\n")
+            if content.endswith("\n"):
+                # split() yields a trailing empty element for the final newline;
+                # drop it so total matches the large-file path (line-by-line
+                # iteration) and the indexer (default_file_chunker), both of
+                # which do not count that trailing newline as its own line.
+                all_lines.pop()
             total = len(all_lines)
             bounds = self._resolve_range(total, start_line, end_line, target)
             if bounds is None:
--- a/tests/unit/test_crud_steps.py
+++ b/tests/unit/test_crud_steps.py
@@ -703,6 +703,39 @@ def test_read_start_line_exceeds_total():
     _run(run())
 
 
+def test_read_trailing_newline_does_not_inflate_total_lines():
+    """A trailing '\\n' must not count as an extra line (small-file path).
+
+    ``line1\\nline2\\nline3\\n`` is 3 lines, matching both the large-file
+    path (line-by-line iteration) and the indexer's chunker. Before the fix,
+    ``content.split("\\n")`` counted the trailing empty element too, so the
+    boundary message reported 4 and start_line=4 (one past real EOF) was
+    silently accepted instead of rejected.
+    """
+
+    async def run():
+        with tempfile.TemporaryDirectory() as tmp, temp_chdir(tmp):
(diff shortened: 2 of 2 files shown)


jinliyl (approved):
Thank you for the clear fix and the focused regression test. The change correctly aligns the small-file line count with the streaming path. LGTM!

jinliyl:
@imrewce code review~

PR #390, opened 2026-07-24
feat(faiss_file_store): upgrade FAISS to HNSW index with async reindex

xyf2020:
## key

核心是优化vector search，把faiss的flat搜索改成了HNSW。但是这会带来一个问题，HNSW索引构建速度很慢。原来代码中，reindex通常在无法正确读取、删除元素个数达到阈值等时候被触发同步执行。使用HNSW后，我将墓碑机制触发reindex通过开关改成了异步执行，reindex完毕之前依然可以访问old index，直到新的index完全构建再和old swap。异步开关默认关闭，与之前的行为对齐并向后兼容。需要注意的是，保证任意时刻只有一个reindex工作在执行，因为reindex时消耗大量CPU资源，并发reindex既没有意义也浪费资源。在异步reindex模式下，使用协程监视flag变量，相当于简化版本的生产者消费者模式，只不过生产者队列中会无条件合并，只有True和False两个状态

## 现在存在的问题

如果在reindex过程中close，那么reindex依然会坚持执行，因为父线程难以强制关闭faiss.add。naive的方法是reindex时每次add一个vector，add一次查看一下thread flag，这样虽然可以及时终止，但是会带来大量的性能下降，HNSW批量插入速度远高于分batch插入，所以权衡之下保留了这个问题。

## Changes
- Switch vector_search to IndexHNSWFlat (inner product + L2 normalize = cosine), with efSearch set to limit*5 at query time.
- Add a tombstone-based logical delete, since HNSW has no physical removal: deletes/updates mark rows dead, search filters them out, and the chunk JSONL stays the source of truth.
- Add an `async_reindex` switch (default off) that keeps the original synchronous rebuild behavior for full backward compatibility.
- When enabled, a rebuild triggered by the tombstone threshold or by embedding backfill runs asynchronously: a new index is built from a snapshot while the old index keeps serving searches, then atomically swapped in.
- The load-time rebuild always stays synchronous regardless of the switch, since there is no old index to serve.
- A single long-lived asyncio coroutine consumes a boolean asyncio.Event flag (repeated submissions coalesce), so exactly one reindex ever runs at a time.
- The CPU-heavy HNSW build is offloaded to a worker thread via asyncio.to_thread (FAISS releases the GIL during add).
- Writes landing during a build bump a counter that re-arms the flag after the swap, folding them into a follow-up rebuild for eventual consistency without lost writes.
- Persist the index as a binary file plus an id-map sidecar, validating index type and dimension on load and rebuilding from chunks on any mismatch.


jinliyl:
Thank you for the detailed description and for adding focused consistency tests. The overall direction makes sense: moving from `IndexFlatIP` to HNSW, keeping the chunk JSONL as the source of truth, and making expensive compaction optionally asynchronous are all reasonable improvements.

I also checked the small `LocalFileStore` change. Extracting `_delete_nodes(nodes)` preserves the previous `LocalFileStore.delete()` behavior. It lets `FaissLocalFileStore` reuse the already-resolved nodes instead of querying the graph twice, which looks good and should reduce the chance of inconsistent results between two lookups.

While reviewing and testing commit `2422109de7505341f8f3288edbc4021974a26f4a`, I found two issues that I think should be addressed before merging.

### 1. An updated embedding can be permanently replaced by a stale snapshot when the store closes during the follow-up rebuild

In asynchronous mode, `_reindex_async()` builds from a snapshot and swaps that snapshot into the live state. If writes occurred during the build, `_index_writes` causes the event to be set for a follow-up rebuild. This provides eventual consistency while the application continues running, but there is a shutdown window:

1. Chunk `a` initially contains text/embedding `alpha`.
2. An asynchronous rebuild starts and snapshots `alpha`.
3. The same chunk ID is updated to text/embedding `beta` while the first build is blocked.
4. The first build finishes and swaps the stale `alpha` snapshot into `_faiss_index`.
5. The worker starts the follow-up rebuild for `beta`.
6. `close()` runs before that follow-up rebuild swaps its result. It cancels the worker and the final `dump()` persists the currently active, stale `alpha` index.

I reproduced this locally. After closing the store, the authoritative chunk JSONL contained `text == "beta"`, while the persisted FAISS index still contained the old `alpha` vector. Searching the persisted index with the beta vector produced a score of `0.0` instead of `1.0`.

This does not self-heal on restart for a same-ID update. `_try_load_sidecar()` validates the index type, dimension, row count, and set of live IDs, but it does not validate that each indexed vector corresponds to the current chunk embedding. Because the ID remains `a`, the stale sidecar is accepted and can remain incorrect across restarts.

Possible ways to address this include:

- reconcile writes that occurred during the build before publishing the new index;
- track an index/chunk generation and avoid writing the sidecar while the active index is behind the authoritative chunks;
- synchronously rebuild on close when a follow-up rebuild is pending; or
- skip/remove the stale sidecar on close so the next load rebuilds it from the chunk JSONL.

Whichever approach is chosen, it would be helpful to add a regression test that updates an existing chunk ID during the first build, waits until the follow-up build starts, closes the store, and then reopens it to verify the persisted vector matches the updated chunk embedding.

### 2. Persisted HNSW indexes do not honor changed construction parameters

The PR introduces `hnsw_m` and `hnsw_ef_construction` as configurable construction parameters, but the persisted sidecar does not record or validate them. `_try_load_sidecar()` accepts any `IndexHNSWFlat` with the expected dimension and live IDs.

For example, I created and persisted an index with `hnsw_m=8`, then reopened the same store with `hnsw_m=32`. The store reported the configured value as 32, but the loaded index still had 16 level-0 neighbors, corresponding to the old M=8 graph. Therefore, changing the configuration is silently ignored until some unrelated event forces a rebuild. The same concern applies to `hnsw_ef_construction`, since it only affects graph construction.

I suggest storing the construction parameters (and ideally an index-format/version marker) in the sidecar metadata and rebuilding from the authoritative chunks when they differ from the active configuration. A restart test with a changed `hnsw_m` would cover this behavior.

For completeness, all GitHub checks are currently green, and the focused test file passed locally (`37 passed`). The existing tests cover normal asynchronous convergence and cancellation, but they do not cover the shutdown window above or reopening with changed HNSW construction parameters.

Thanks again for the work on this. Once these persistence cases are handled, the implementation should be much safer while preserving the local-first/source-of-truth guarantees.

reme/components/file_store/local_file_store.py
```diff
@@ -450,6 +450,15 @@ async def dump(self) -> None:
             await self.keyword_index.dump()
         await self.file_graph.dump()
 
+    # -- maintenance -----------------------------------------------------------
+
```
jinliyl, line 454: optimize index

reme/config/default.yaml
```diff
@@ -624,20 +630,20 @@ components:
...
-#      credential:
-#        api_key: ${EMBEDDING_API_KEY:-}
-#        base_url: ${EMBEDDING_BASE_URL:-https://dashscope.aliyuncs.com/compatible-mode/v1}
-#      parameters: { }
-#
-#  embedding_store:
-#    default:
-#      backend: local
-#      as_embedding: default
+  as_embedding:
```
jinliyl: 还是注释

reme/steps/index/refine_store.py
```diff
@@ -0,0 +1,16 @@
+"""Idle-time file store maintenance, scheduled off the request path (e.g. cron)."""
```
jinliyl: 名字改下

reme/config/default.yaml
```diff
@@ -68,6 +68,12 @@ jobs:
       - backend: dream_finish_step
         file_catalog: dream
 
+  refine_store_cron:
```
jinliyl: 名字改掉

reme/config/default.yaml
```diff
@@ -727,9 +733,8 @@ components:
 
   file_store:
     default:
-      backend: local
+      backend: faiss
```
jinliyl: 注释

pyproject.toml
```diff
@@ -23,6 +23,7 @@ classifiers = [
 dependencies = [
     "aiofiles>=24.1.0",
     "croniter>=2.0",
+    "faiss-cpu>=1.13.2",
```
jinliyl: 还原下

jinliyl:
感谢继续更新这个 PR。最新提交中，异步重建检测到并发写入后会丢弃旧快照，不再把旧快照切换为当前索引；此前提到的那个具体竞态已经得到修复。相关测试和当前 GitHub CI 也都通过了。

我进一步复核持久化和配置变更后，仍发现以下几个边界情况，建议合并前再处理一下：

1. **同 ID、同维度的旧向量 sidecar 仍可能被接受。** `_try_load_sidecar()` 目前只校验索引类型、维度、行数和 live ID 集合，没有验证每个 ID 对应的向量是否与 chunk JSONL 中的 embedding 属于同一代。由于 `dump()` 是先写权威 chunk JSONL、再写 FAISS sidecar，如果进程在两次写入之间退出，重启时就可能接受旧向量。我本地复现时，JSONL 中 chunk 已更新为 `beta`、embedding 为 `[0, 1]`，但加载的索引仍是旧的 `[1, 0]`，用 beta 查询得分为 `0.0`（预期 `1.0`）。建议在 sidecar 中保存 chunk/embedding generation 或稳定摘要，不一致时从 JSONL 重建。

2. **空 HNSW 索引会绕过 `hnsw_m` 校验。** 当前只有 `index.ntotal > 0` 时才通过 `nb_neighbors(0)` 反推并校验 M。我实测先以 `M=8` 保存空索引，再以 `M=32` 重开，sidecar 会被接受，但实际加载的图仍是 `M=8`，后续新增向量也会继续使用旧结构。建议把 M 明确写入 sidecar 元数据，即使索引为空也能校验。

3. **修改 `hnsw_ef_construction` 只影响后续新增向量。** 当前加载时仅更新 `index.hnsw.efConstruction`，现有图的边仍由旧参数构建，因此配置变更后索引主体并没有采用新参数，之后还可能形成混合构建质量的图。既然这是构建参数，建议配置变化时触发一次从权威 chunks 的完整重建，或者明确记录并说明其仅对后续插入生效。

整体方向没有问题，HNSW、墓碑过滤和异步重建的实现也已经完善了不少。把上述持久化代际和构建参数问题补齐后，会更符合 chunk JSONL 作为 source of truth 的约定。辛苦了！

jinliyl (approved):
LGTM

PR #391, opened 2026-07-24
feat(Step tools): add white/black path prefix permission filtering to read, edit, write

xyf2020:
## Summary

Add a reusable `PrefixCheck` mixin that provides white/black path-prefix permission filtering for file I/O steps (`ReadStep`, `WriteStep`, `EditStep`). This enables restricting file access to specific workspace subdirectories via YAML configuration.

### New Module
- `reme/steps/file_io/prefix_check.py`: `PrefixCheck` mixin with two-stage permission check:
  - **White stage** (`white_path_prefix`): allowlist of path prefixes; `None` disables, `[]` denies all
  - **Black stage** (`black_path_prefix`): denylist applied after white stage; `None`/`[]` disables
  - Path-aware matching using `PurePath.is_relative_to()` to respect component boundaries

### Modified Steps
- `ReadStep`, `WriteStep`, `EditStep` now inherit from `PrefixCheck` and call `_check_path_permission()` before file operations

### Configuration
- `default.yaml` / `lme.yaml`: add `read_daily`, `edit_daily`, `write_daily` jobs scoped to `daily/` directory via `white_path_prefix: ["daily/"]`
- `auto_memory.yaml`: update to use the new daily-scoped jobs

### Tests
- `tests/unit/test_path_permission.py`: comprehensive coverage for white/black list combinations, path boundary matching, and edge cases


jinliyl:
Thank you for working on this and for adding the path-boundary tests. The reusable path checking itself is helpful, but the current approach of adding separate `read_daily`, `edit_daily`, and `write_daily` jobs to the configuration is not quite aligned with the direction I had in mind.

I would prefer to keep using the existing `read`, `edit`, `write`, and `frontmatter_update` jobs, and have `AutoMemoryStep` inject a server-owned, request-scoped path constraint when exposing those tools to the agent. For an existing note, the allowed scope could be the exact `note_path` rather than the whole `daily/` directory. The injected constraint should not be visible to or overridable by the model, and the file I/O steps should read it from the per-invocation `RuntimeContext`.

This would avoid duplicating jobs in each config, work with a customized `daily_dir`, keep the BEAM/LME configurations consistent, and also allow us to apply the same restriction to `frontmatter_update` so there is no metadata-update bypass. The creation path could use the same injection mechanism to pin the `date` passed to `daily_write`.

Would you be open to adjusting the implementation in that direction? Thanks again for the contribution.

jinliyl:
Thank you again. To make the suggested direction more concrete, here is a possible implementation plan. I think this can be done without adding `read_daily`, `edit_daily`, or `write_daily` jobs.

### 1. Keep the existing job names

`AutoMemoryStep` can continue to expose the existing jobs:

```python
self.create_tools = ["daily_write"]
self.update_tools = ["read", "edit", "frontmatter_update", "write"]
```

The restriction should belong to one `agent_wrapper.reply()` invocation, rather than to a second set of jobs in YAML.

### 2. Inject server-owned kwargs into every selected job tool

A small wrapper-level option such as `injected_job_kwargs` could be introduced. These kwargs must be added by the wrapper after receiving the model's tool arguments, and the model must not be able to override them.

For the create flow, only `daily_write` is exposed, so we can pin the resolved date:

```python
injected_job_kwargs = {"date": day}
```

For the update flow, all four exposed tools operate on the same existing note, so we can inject an exact path scope:

```python
injected_job_kwargs = {
    "_allowed_paths": [note_path],
}
```

The call in `AutoMemoryStep.execute()` would then be conceptually:

```python
result = await self.agent_wrapper.reply(
    user_message,
    system_prompt=self.prompt_format("system_prompt"),
    job_tools=self.create_tools if created else self.update_tools,
    injected_job_kwargs=(
        {"date": day}
        if created
        else {"_allowed_paths": [note_path]}
    ),
    **self._reply_extra_kwargs(day),
)
```

Using the exact `note_path` is preferable to allowing the entire daily directory: the update prompt already says that only one target note may be changed, and an exact scope provides the least privilege. It also works automatically when `daily_dir` is customized.

The underscore-prefixed `_allowed_paths` is intentionally not part of any public job parameter schema, so it is not presented to the model as a tool argument.

### 3. Apply the injection consistently in all agent wrappers

For the AgentScope and Claude Code wrappers, the generated tool handler can merge the injected values immediately before calling the job:

```python
async def run_job(**model_kwargs):
    conflicts = model_kwargs.keys() & injected_job_kwargs.keys()
    if conflicts:
        raise ValueError("injected tool arguments cannot be provided by the model")

    call_kwargs = {**model_kwargs, **injected_job_kwargs}
    response = await job(**call_kwargs)
    ...
```

Rejecting conflicts is useful because it makes the ownership boundary explicit. Silently allowing model values to win would make the restriction bypassable.

For the Codex wrapper, the same mapping needs to be serialized to the dedicated MCP subprocess and passed into `MCPService`. `MCPService` already has an `injected_job_kwargs` mechanism, so the existing Codex `tool_context_id` injection path can be generalized to carry these additional values as well. The injected keys should remain omitted from the exposed MCP tool schema.

The existing `tool_defaults` argument used by the BEAM/LME subclasses does not currently appear to be consumed by the wrappers. It could either be replaced by `injected_job_kwargs`, or implemented with the same server-owned semantics. I slightly prefer the latter name because these values are enforced constraints, not ordinary defaults that callers may override.

### 4. Read the scope from `RuntimeContext`, not from mutable Step state

`BaseJob` already creates fresh Step instances per invocation, and injected job kwargs become request-scoped `RuntimeContext` values. The path-checking helper can therefore read `_allowed_paths` from `self.context` during `execute()`.

A check could normalize both the requested target and every allowed path through the existing workspace-safe path resolver, then compare the resolved paths:

```python
def _check_runtime_path_scope(self, target: Path) -> bool:
    assert self.context is not None
    raw_allowed = self.context.get("_allowed_paths")
    if raw_allowed is None:
        return True

    allowed: list[Path] = []
    for raw in raw_allowed:
        resolved, err = resolve_path(self.workspace_path, str(raw))
        if err or resolved is None:
            return False  # invalid permission configuration should fail closed
        allowed.append(resolved)

    return target.resolve() in allowed
```

This is only illustrative; it may be cleaner to incorporate exact-path and prefix checks into the proposed `PrefixCheck` helper. The important properties are:

- the scope is request-local;
- all comparisons use resolved workspace-contained paths;
- invalid injected constraints fail closed;
- no mutable permission state is retained on a Step between calls.

### 5. Cover all mutating tools exposed to the agent

The check should run in:

- `ReadStep`
- `EditStep`
- `WriteStep`
- `FrontmatterUpdateStep`

Including `FrontmatterUpdateStep` is important; otherwise the agent can still modify frontmatter on an arbitrary workspace Markdown file even when read/edit/write are restricted.

Each Step should resolve/gate the target first, then run the permission check before reading or writing the file.

### 6. Remove the duplicated YAML jobs and prompt renames

With request-scoped injection in place, the following changes would no longer be needed:

- `read_daily`, `edit_daily`, and `write_daily` in `default.yaml`;
- the same jobs in `lme.yaml`;
- prompt changes from `read/edit/write` to `read_daily/edit_daily/write_daily`;
- changing the shared `AutoMemoryStep.update_tools` list to the new job names.

This also avoids breaking `beam.yaml`, which currently defines only the original jobs.

### 7. Suggested focused tests

I would suggest adding tests for these observable behaviors:

1. An AutoMemory update can read/edit/write its exact `note_path`.
2. The same tools reject another note in the same daily directory.
3. They reject a path outside the daily directory.
4. `frontmatter_update` rejects a different Markdown file.
5. A customized `daily_dir`, for example `journal`, works without adding config-specific jobs.
6. AgentScope, Claude Code, and Codex wrappers inject the hidden kwargs and reject model-supplied conflicts.
7. Invalid injected path constraints fail closed.
8. BEAM and LME continue resolving the original tool names.

This should keep the configuration small while making the permission boundary narrower and consistent across agent backends. I hope this clarifies the proposed direction, and I am happy to discuss naming or implementation details. Thanks again for taking this on.

reme/steps/benchmark/beam/auto_memory.py
```diff
@@ -117,7 +117,11 @@ def _interpolate_timestamps(items: list[dict]) -> list[dict]:
```
jinliyl, line 117: auto-memory不需要修改

reme/steps/file_io/edit.py
```diff
@@ -5,18 +5,22 @@
...
 from ._file_io import get_path_lock, read_file_safe, write_file_safe
 from ._path import NON_MD_WARNING, gate_md, resolve_path
+from .prefix_check import PrefixCheck
 from ..base_step import BaseStep
 from ...components import R
 
 
 @R.register("edit_step")
-class EditStep(BaseStep):
+class EditStep(PrefixCheck, BaseStep):
```
jinliyl: 不要双继承

reme/steps/file_io/_path.py
```diff
@@ -82,10 +82,11 @@ def resolve_path(
             return workspace_path.resolve(), None
         return None, "`path` is required"
     s = str(raw).strip()
-    p = Path(s)
+    p = Path(s).expanduser()
```
jinliyl: mark下

jinliyl (approved):
LGTM

PR #406, opened 2026-07-30
feat(evaluation): track job calls and agent token usage in benchmarks

xyf2020:
## Summary

- Add application-lifetime counters for job invocations.
- Add a backend-neutral `TokenUsage` schema and record usage from AgentScope,
  Claude Code, and Codex agent wrappers.
- Add evaluation helpers for reading and tracking job-count and token-usage deltas.
- Report agentic search-call and token-usage mean/variance in BEAM and LongMemEval.

## Notes

- Counters are in-memory application metadata and remain rebuildable/non-durable.
- Cache and reasoning metrics are optional because backend reporting differs.

## Validation

- `pytest tests/unit/test_evaluation_interface.py tests/unit/test_token_usage.py tests/unit/test_job.py tests/unit/test_utils.py tests/unit/test_cc_agent_wrapper.py tests/unit/test_codex_agent_wrapper.py -v`


jinliyl:
The job-call accounting can be simplified substantially without losing any behavior required by the evaluation helpers.

At the moment, `_counter_key()` builds a key from both the built-in entry class and every project-specific subclass in the MRO, for example:

```python
["__job_counter", "StreamJob", "ProjectStreamJob", "chat"]
```

`check_job_count()` then has to reconstruct the same path by importing all four built-in job implementations, walking the concrete job's MRO, selecting an entry class, and calling the protected `_counter_key()` method.

However, the observable API only asks one question: how many times did the registered job named `job_name` run in this `ApplicationContext`? Job names are already unique within `app_context.jobs`, and none of the current benchmark consumers aggregate or query counts by implementation class or subclass. Consequently, the class and MRO segments do not add information used by a reader; they only make the writer, reader, and tests depend on the same internal inheritance details.

I suggest storing job counts directly by registered job name:

```python
JOB_COUNTER_PREFIX = "__job_counter"


def _record_call(self) -> None:
    """Increment this job's application-lifetime invocation count."""
    metadata = getattr(self.app_context, "metadata", None)
    if isinstance(metadata, dict):
        global_counter_inc(metadata, [JOB_COUNTER_PREFIX, self.name])
```

Each execution entry point would then use the same call:

```python
self._record_call()
```

The read side becomes correspondingly smaller:

```python
def check_job_count(job_name: str, app_context: "ApplicationContext") -> int:
    if job_name not in app_context.jobs:
        raise KeyError(f"Job '{job_name}' not found")
    return global_counter_get(
        app_context.metadata,
        [JOB_COUNTER_PREFIX, job_name],
    )
```

This would remove:

- `BaseJob._counter_key()`;
- the `entry_class` argument to `_record_call()`;
- `_JOB_ENTRY_CLASSES`;
- the `BackgroundJob`, `CronJob`, and `StreamJob` imports from `evaluation_interface.py`;
- the MRO traversal and its unsupported-entry-class branch;
- tests coupled specifically to the serialized subclass path.

It also makes custom subclasses work automatically: a `ProjectStreamJob(name="chat")` records and reads `chat` exactly like any other registered job, without encoding Python class names into application metadata. This is useful because class names and inheritance structure are implementation details, while the registered job name is the stable application-level identity.

I would still keep the one-line `_record_call()` invocation in each concrete execution path rather than moving all accounting into `Application.run_job()`. `BackgroundJob` supervisor retries and individual `CronJob` triggers can execute outside a normal user-facing `run_job()` call, so centralizing only at the application dispatch layer would undercount those executions. The proposed simplification changes the counter key, not the points at which calls are counted.

Since this counter layout is introduced by this PR and has not yet become a released contract, this is the best time to simplify it without requiring a metadata compatibility path. The resulting behavior remains the same for BEAM, LongMemEval, and embedded consumers such as QwenPaw: counters are still application-scoped, in-memory, thread-safe, and keyed by the job identity that callers already use.

jinliyl:
Thanks for adding the job-call and token-usage statistics. I took another look specifically at the streaming versus non-streaming paths and at whether the implementation can be reduced in scope.

For the current BEAM and LongMemEval use case, the benchmark calls `app.run_job(...)` without a `stream_queue`. Therefore, `RuntimeContext.stream` is `False`, and `BaseAgenticAnswerStep.execute()` calls `agent_wrapper.reply(...)`; it does not call `reply_stream(...)`.

Given that scope, I think it would be reasonable—and preferable—to keep token accounting in the non-streaming `reply()` path only for this PR. The existing `ChunkEnum.USAGE` events in the streaming APIs should remain intact for compatibility, but the new application-level counter updates and usage-format changes in `reply_stream()` could be reverted. This keeps the change focused on the benchmark requirement and avoids introducing streaming accounting semantics that are not yet needed.

Streaming accounting also needs a little more design before it can be considered reliable. In particular, consumers may stop iterating early or close an async generator. The Codex implementation currently records `final_usage` in a `finally` block, so a partially consumed or interrupted turn can be counted even though `_record_token_usage()` describes a completed invocation. AgentScope records only after normal stream exhaustion, while Claude Code records when it receives a terminal result. Those backend-specific differences make the meaning of a streaming counter inconsistent. It would be cleaner to address cancellation, interruption, failure, and partial-consumption semantics together in a separate change when streaming statistics are actually required.

The non-streaming implementations can also be simplified:

1. **AgentScope**

   The current `reply()` implementation switches from `await agent.reply()` to consuming `agent.reply_stream()` solely to collect `ModelCallEndEvent` objects. AgentScope already accumulates the input/output usage of all model calls belonging to the same reply into the final assistant message usage. We can therefore preserve the normal non-streaming execution path and read `last_msg.usage`:

   ```python
   await agent.observe(inputs)
   last_msg = await agent.reply()

   usages: list[TokenUsage] = []
   if last_msg.usage is not None:
       usages.append(
           TokenUsage(
               input_tokens=last_msg.usage.input_tokens,
               output_tokens=last_msg.usage.output_tokens,
           ),
       )
   ```

   If structured output triggers an additional model call, its normalized `res.usage` can still be appended, followed by `TokenUsage.combine(usages)`. This avoids changing the non-streaming execution path just for instrumentation.

   There is still an AgentScope/Anthropic limitation that should be handled explicitly: `ModelCallEndEvent` and the persisted message `Usage` expose only input/output totals and discard cache dimensions. AgentScope Anthropic keeps `input_tokens`, cache-read tokens, and cache-creation tokens separately at the model-response layer. Consequently, treating the event/message `input_tokens` as the complete prompt size undercounts Anthropic input and total tokens whenever caching is used. The new `_agentscope_usage()` test validates normalization from a full `ChatUsage`, but the main agent reply path never receives that full object. If the complete provider usage cannot be accessed here, I suggest reporting the cache metrics as unavailable and documenting the input-token limitation rather than claiming a cache-inclusive value.

2. **Claude Code**

   The result usage is normalized, dumped, and then parsed back into the same model before recording. A single local variable is sufficient:

   ```python
   usage = self._claude_usage(last_msg.usage)
   result = {
       # existing fields...
       "usage": usage.model_dump(),
   }
   self._record_token_usage(usage)
   ```

   This removes the unnecessary `TokenUsage.model_validate(result["usage"])` round trip.

3. **Codex**

   The non-streaming branch can similarly normalize once and reuse the value for both the response and the counter:

   ```python
   raw_usage = getattr(result, "usage", None)
   usage = self._codex_usage(raw_usage.last) if raw_usage is not None else TokenUsage()
   response["usage"] = usage.model_dump()
   if raw_usage is not None:
       self._record_token_usage(usage)
   ```

My suggested scope for this PR would therefore be:

- record application-level token usage from non-streaming `reply()` calls;
- let the benchmark trackers read deltas from those counters;
- preserve the pre-existing streaming usage chunks without adding stream-level counters yet;
- add a focused test showing that the benchmark path uses `reply()` and that AgentScope final-message usage includes all ReAct model iterations;
- explicitly document or resolve the AgentScope/Anthropic cache-token limitation.

This would make the implementation smaller and easier to reason about while fully satisfying the current benchmark requirement. Streaming counter support can then be added later with one clearly defined cross-backend contract for successful completion, cancellation, and partial consumption.

jinliyl:
感谢补充 AgentScope 双版本兼容逻辑。我重新核对了 AgentScope `v2.0.4.post1` 和 `v2.0.5` 的正式 tag，关于 usage 这里有一点需要更正：`Msg.usage` 以及整个 reply 内多次模型调用的累计，并不是 2.0.5 才引入的。

具体证据如下：

- `v2.0.4.post1` 已经在 `Msg` 上定义了 `usage: Usage | None`：
  https://github.com/agentscope-ai/agentscope/blob/v2.0.4.post1/src/agentscope/message/_base.py#L57-L85
- `v2.0.4.post1` 每收到一次 `ModelCallEndEvent`，都会初始化或累加 `Msg.usage.input_tokens/output_tokens`：
  https://github.com/agentscope-ai/agentscope/blob/v2.0.4.post1/src/agentscope/message/_base.py#L238-L250
- `v2.0.4.post1` 构造最终回复消息时，会把上下文中已经累计的 usage 复制到最终 `AssistantMsg`：
  https://github.com/agentscope-ai/agentscope/blob/v2.0.4.post1/src/agentscope/agent/_agent.py#L1101-L1115
- `v2.0.4.post1` 的官方测试也明确验证了两次 `ModelCallEndEvent` 的累计结果：`10/20 + 5/8 = 15/28`：
  https://github.com/agentscope-ai/agentscope/blob/v2.0.4.post1/tests/event_to_message_test.py#L676-L714
- `v2.0.5` 保留了相同的累计机制：
  https://github.com/agentscope-ai/agentscope/blob/v2.0.5/src/agentscope/message/_base.py#L278-L286

因此，当前注释中“token usage 从逐次 `ModelCallEndEvent` 移到 2.0.5 的聚合 `Msg.usage`”这一前提并不准确；两个版本实际上同时具有逐次事件和最终消息累计 usage。

建议两个版本统一保留正常的非流式调用路径：

```python
await agent.observe(inputs)
last_msg = await agent.reply()
usage = (
    TokenUsage.from_provider(last_msg.usage)
    if last_msg.usage is not None
    else None
)
```

这样可以删除 usage 获取部分的 `<2.0.5 / >=2.0.5` 分支，也不需要为了统计而把 2.0.4.post1 的非流式调用改成消费 `reply_stream()`。如果后续还有 structured output 的额外模型调用，继续将 `res.usage` 与这里的 usage 合并即可。

另外，两个版本的最终 `Msg.usage` 和 `ModelCallEndEvent` 都只提供 input/output tokens，不包含 cache/reasoning 维度；统一使用最终消息不会比收集事件损失更多信息。

jinliyl (approved):
LGTM

PR #408, opened 2026-07-31
docs: fix ReMe documentation links

ployts:
## Summary

- point the README documentation entry to `https://docs.agentscope.io/reme`
- update English and Chinese contribution-guide links to the current `latest/.../contribution` routes
- add an official documentation badge to both READMEs

## Validation

- `git diff --check`
- verified the documentation root and both contribution links return HTTP 200

Tests were not run because this change only updates documentation links.


jinliyl (approved):
LGTM

PR #410, opened 2026-08-03
feat(file_store): add ZvecLocalFileStore backend

chinaux:
## Summary

This PR adds a new `ZvecLocalFileStore` backend that uses [zvec](https://zvec.org) as an in-process vector index for ANN search, alongside the existing `LocalFileStore` (brute-force) and `FaissLocalFileStore` (FAISS/HNSW) backends.

## Motivation

zvec is an embedded vector database that requires no separate service process, making it attractive for personal/local deployments where running FAISS or an external vector DB is unnecessary. The new backend follows the same architecture as `FaissLocalFileStore`: chunk metadata and text remain authoritative in the compressed JSONL, while zvec stores only `(chunk_id, embedding)` pairs for fast approximate nearest-neighbor search.

## What changed

- **New backend**: `reme/components/file_store/zvec_local_file_store.py`
  - Native zvec `Collection.upsert` / `Collection.delete` (no tombstone bookkeeping).
  - HNSW index with configurable `hnsw_m` and `hnsw_ef_construction`.
  - JSON digest sidecar binds the collection to the chunk generation; mismatches (crash, model change, external tampering) trigger a clean rebuild from `self.file_chunks`.
  - Dimension and HNSW `M` are validated on load; mismatches also rebuild.
  - `optimize_index()` delegates to zvec's native `Collection.optimize()`.

- **Registration**: `ZvecLocalFileStore` is registered under the `"zvec"` name in `reme/components/file_store/__init__.py`.

- **Tests**:
  - Added `tests/unit/test_zvec_file_store.py` (14 dedicated tests for CRUD, persistence, rebuild triggers, backfill, optimize).
  - Parametrized `tests/unit/test_file_store_consistency.py` so existing consistency tests run against both `LocalFileStore` and `ZvecLocalFileStore`.

## Test results

```bash
pytest tests/unit/test_zvec_file_store.py tests/unit/test_file_store_consistency.py -q
# 63 passed in ~3.5s
```

```bash
pre-commit run --files \
  reme/components/file_store/zvec_local_file_store.py \
  tests/unit/test_zvec_file_store.py \
  tests/unit/test_file_store_consistency.py \
  reme/components/file_store/__init__.py
# all hooks passed
```

## Compatibility and impact

- Existing workspaces are unaffected unless the user explicitly switches to the `zvec` backend.
- When a zvec collection is opened, its vector dimension and HNSW `M` are validated against the current configuration; any mismatch rebuilds the collection from the JSONL chunks automatically.
- Keyword search is unchanged and still provided by the inherited BM25 keyword index.

## Known limitations

1. Dynamic reassignment of `embedding_store` after `start()` will not automatically materialize a zvec collection. This mirrors the existing `FaissLocalFileStore` behavior, where the backend assumes components are resolved at startup.
2. Changing `hnsw_ef_construction` on an existing store does not trigger a rebuild (only `hnsw_m` is validated), because zvec bakes `ef_construction` into the graph at build time and does not expose runtime updates like FAISS does.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,6 +48,7 @@ core = [
     "openai-codex>=0.144.4",
     "pproxy>=2.7.9,<3",
     "faiss-cpu>=1.13.2",
+    "zvec>=0.6.0",
     "jieba>=0.42.1",
     "rjieba>=0.2.1",
     "neo4j>=6.2.0",
--- a/reme/components/file_store/__init__.py
+++ b/reme/components/file_store/__init__.py
@@ -8,9 +8,11 @@ shipped today.
 from .base_file_store import BaseFileStore
 from .faiss_local_file_store import FaissLocalFileStore
 from .local_file_store import LocalFileStore
+from .zvec_local_file_store import ZvecLocalFileStore
 
 __all__ = [
     "BaseFileStore",
     "FaissLocalFileStore",
     "LocalFileStore",
+    "ZvecLocalFileStore",
 ]
--- a/reme/components/file_store/zvec_local_file_store.py
+++ b/reme/components/file_store/zvec_local_file_store.py
@@ -3,6 +3,7 @@
 import hashlib
 import json
 import shutil
+import time
 from uuid import uuid4
 
 import aiofiles
@@ -14,6 +15,8 @@ from ...schema import FileChunk, FileNode
 
 # Batch size for bulk inserts during a rebuild.
 _ZVEC_INSERT_BATCH_SIZE = 1024
+# Batch size for the startup fetch that verifies persisted collection contents.
+_ZVEC_VERIFY_BATCH_SIZE = 1024
 
 
 @R.register("zvec")
@@ -33,7 +36,9 @@ class ZvecLocalFileStore(LocalFileStore):
     sidecar records an order-independent digest of the live
     ``(chunk_id, embedding)`` set at dump time; on load a digest mismatch
     (crash between the chunk dump and the collection flush, externally mixed
-    files, model change) triggers a clean rebuild.
+    files, model change) triggers a clean rebuild, and the opened collection is
+    additionally verified against the ids and vectors it should hold so a
+    damaged or externally modified collection is rebuilt instead of served.
 
     HNSW parameters (``hnsw_m``, ``hnsw_ef_construction``) are baked into the
     collection schema at creation time; changing ``hnsw_m`` on an existing
@@ -43,7 +48,8 @@ class ZvecLocalFileStore(LocalFileStore):
     idle-time index compaction hook.
 
     zvec is imported lazily inside ``__init__`` so that merely importing this
-    module does not require the optional dependency.
+    module does not require the optional dependency; the backend targets the
+    zvec version declared in ``pyproject.toml`` (``zvec>=0.6.0``).
     """
 
     def __init__(
@@ -67,7 +73,7 @@ class ZvecLocalFileStore(LocalFileStore):
             import zvec
         except ImportError as e:
             raise ImportError(
-                "zvec is required for ZvecLocalFileStore. Install with `pip install zvec`.",
+                "zvec is required for ZvecLocalFileStore. Install with `pip install 'zvec>=0.6.0'`.",
             ) from e
         return zvec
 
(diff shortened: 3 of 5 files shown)


jinliyl:
Thanks for adding the zvec backend. I ran the focused tests with `zvec==0.6.0` and found several issues that should be addressed before merging.

## 1. The collection is not updated when an explicit embedding changes without a text change

**Severity: High — data consistency**

In `ZvecLocalFileStore._sync_collection_after_upsert()`, the following condition assumes that unchanged text means an unchanged embedding:

```python
if cid in self._indexed_ids and old_text_by_id.get(cid) == chunk.text:
    continue
```

That assumption does not hold because `LocalFileStore` explicitly accepts a caller-provided embedding when it has the expected dimension.

I reproduced this sequence:

1. Upsert chunk `same` with text `"same text"` and embedding `[1, 0]`.
2. Upsert the same chunk ID and text again, explicitly providing `[0, 1]`.
3. `self.file_chunks["same"].embedding` becomes `[0, 1]`.
4. The condition above skips the zvec upsert, so the collection still contains `[1, 0]`.

A query matching `[0, 1]` continued to return a similarity score of `0.0` instead of `1.0`.

The inconsistency also survives restart. `dump()` calculates the sidecar digest from the new embedding in `self.file_chunks`, while the collection still contains the old vector. On the next startup, the digest matches the JSONL chunks, so `_try_open_collection()` accepts the stale collection without rebuilding it.

Suggested fix: compare the old and new embeddings, have the parent upsert path expose whether the exact old embedding was reused, or conservatively upsert every eligible chunk included in the request. Please add a regression test for a same-ID, same-text update with a different valid explicit embedding, including a close/reopen round trip.

## 2. The sidecar validation does not verify the actual collection contents

**Severity: High — stale or damaged indexes may be accepted**

`_try_open_collection()` compares the sidecar digest and ID list against `self.file_chunks`, then validates the collection schema. It does not verify that the IDs or vectors recorded in the sidecar are actually present in the opened zvec collection.

I reproduced this case:

1. Create a store containing chunk `a` and close it normally.
2. Open the zvec collection directly, delete document `a`, and flush it without modifying the sidecar.
3. Start `ZvecLocalFileStore` again.

The store reports that it opened one vector, but the actual state is:

```text
_indexed_ids = {"a"}
collection.stats.doc_count = 0
vector_search(...) = []
```

Therefore, the sidecar is bound to the authoritative JSONL, but not to the zvec collection itself. This means collection corruption or external tampering can be accepted as valid.

At minimum, startup should verify:

```python
collection.stats.doc_count == len(expected_ids)
```

That catches missing or extra documents. To detect a different collection with the same document count, the implementation should also validate the actual collection IDs or store a generation marker inside the collection. If same-ID vector corruption must be detected, the actual stored vectors need to be covered by the validation mechanism as well.

Please add tests for missing and unexpected collection documents, and ideally for an expected ID containing a different vector.

## 3. `zvec` is not declared as a project dependency

**Severity: High — the backend has no supported installation path**

The PR registers a `"zvec"` backend but does not update `pyproject.toml`. Neither documented installation command currently installs zvec:

```bash
pip install "reme-ai[core]"
pip install "reme-ai[full]"
```

The new tests call `pytest.skip()` when zvec is unavailable, so CI can remain green without executing the new backend behavior.

Since `faiss-cpu` is already part of the `core` extra and zvec serves the same category of optional vector-index backend, I suggest adding zvec there with an explicit supported version range, for example:

```toml
core = [
    ...
    "faiss-cpu>=1.13.2",
    "zvec>=0.6.0",
]
```

This will also ensure that the unit-test and pre-commit workflows, which install `.[dev,core]`, actually execute the zvec tests. The supported Python and platform wheel matrix should be confirmed before choosing the final constraint.

This does not affect the current QwenPaw integration: QwenPaw depends on the base `reme-ai` package rather than `reme-ai[core]`, and its embedded configuration explicitly uses the `"local"` file-store backend.

## 4. The implementation uses an already-deprecated zvec API

**Severity: Medium — forward compatibility**

`_query_collection()` currently uses:

```python
self._zvec.VectorQuery(field_name="embedding", vector=vector)
```

With `zvec==0.6.0`, `VectorQuery` is a deprecated alias and emits:

```text
DeprecationWarning: VectorQuery is deprecated and will be removed in a future version. Use Query instead.
```

The new backend should use the current API:

```python
self._zvec.Query(field_name="embedding", vector=vector)
```

## Validation

I ran:

```bash
pytest tests/unit/test_zvec_file_store.py \
       tests/unit/test_file_store_consistency.py -q
```

Result:

```text
63 passed, 20 warnings
```

The passing tests do not cover the two consistency failures above. The warnings are caused by the deprecated `VectorQuery` API.

chinaux:
Thanks for the review — All four points are fixed (upsert no longer assumes unchanged text means an unchanged vector, the collection's doc_count/ids/vectors are validated on open, zvec>=0.6.0 is declared in the core extra, VectorQuery → Query), with six regression tests added — 69 passed, no warnings.

jinliyl (approved):
LGTM

PR #424, opened 2026-08-06
fix(prompt): default omitted conditional flags to false

RerankerGuo:
## Summary

- Always apply conditional-line filtering so `[flag]` lines are omitted unless the corresponding boolean flag is explicitly true.
- Add regression coverage for omitted flags, including templates that also interpolate format variables.

## Problem

`PromptHandler` documents conditional lines as opt-in, but `prompt_format()` skipped filtering when no boolean kwargs were supplied. A template such as `[debug] secret detail` therefore leaked the raw tagged line into the rendered prompt by default.

## Verification

- Before: `PromptHandler(p="[debug] secret detail\nbase").prompt_format("p")` returned `"[debug] secret detail\nbase"`.
- After: the same reproduction returns `"base"`.
- `pytest tests/unit/test_prompt_handler.py -q` -> `32 passed`.
- `pre-commit run --files reme/components/prompt_handler.py tests/unit/test_prompt_handler.py` -> all hooks passed.

## Scope

- No prompt syntax or public method signatures change.
- Explicit `flag=True` behavior is unchanged.
- Unflagged lines and normal format variables are unchanged.

### DIFF
--- a/reme/components/prompt_handler.py
+++ b/reme/components/prompt_handler.py
@@ -116,8 +116,7 @@ class PromptHandler:
         flags = {k: v for k, v in kwargs.items() if isinstance(v, bool)}
         formats = {k: v for k, v in kwargs.items() if not isinstance(v, bool)}
 
-        if flags:
-            prompt = self._apply_flag_filter(prompt, flags)
+        prompt = self._apply_flag_filter(prompt, flags)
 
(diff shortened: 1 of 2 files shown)


jinliyl (approved):
LGTM

PR #426, opened 2026-08-06
fix(component): roll back failed startup

RerankerGuo:
## Summary

Roll back partially started component lifecycles when `BaseComponent.start()` fails.

## Problem

`start()` currently starts owned dependencies before invoking the parent's `_start()`. If the parent hook raises, the parent remains marked as not started, so a later `close()` returns immediately while the owned components remain running.

This can leak clients, tasks, or other resources created by default-factory dependencies.

## Changes

- Track owned components that successfully started during the current attempt.
- If parent startup fails, invoke its cleanup hook and close started owned components in reverse order.
- Preserve the original startup exception even if rollback cleanup also fails.
- Add a regression test covering failed parent startup and defensive follow-up close.

## Evidence

Before the fix:

```text
FAILED test_start_failure_rolls_back_parent_and_owned_components
assert parent.close_count == 1
E assert 0 == 1
```

After the fix:

```text
pytest tests/unit/test_base_component.py -q
25 passed

pytest tests/unit/test_job.py -q
21 passed
```

All relevant pre-commit hooks pass.

## Scope

Successful startup and normal idempotent close behavior are unchanged.

### DIFF
--- a/reme/components/base_component.py
+++ b/reme/components/base_component.py
@@ -210,10 +210,28 @@ class BaseComponent(ComponentMixin, ABC):
             if self._is_started:
                 return
             await self._resolve_bindings()
-            for owned in self._owned:
-                await owned.start()
-            await self._start()
-            self._is_started = True
+            started_owned: list[BaseComponent] = []
+            parent_start_attempted = False
+            try:
+                for owned in self._owned:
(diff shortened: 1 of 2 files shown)


jinliyl:
LGTM

Issue #429, opened 2026-08-06
Chinese frontmatter description not searchable via BM25; `stat`/`frontmatter_read`/`frontmatter_update` do not auto-append `.md` suffix


chinaux:
## Description

I noticed two related inconsistencies while following the quick-start docs (`docs/zh/quick_start.md` and `docs/en/quick_start.md`).

### Bug 1: Chinese frontmatter description returns no search results

After creating a note with a Chinese `description` as shown in the docs:

```markdown
---
name: Quick Start Demo
description: "快速开始示例记忆"
---
# Quick Start Demo
```

The recommended Chinese query returns nothing:

```bash
reme search query="快速开始 示例 记忆" limit=5
```

Whereas the English query works as expected:

```bash
reme search query="quick start example memory" limit=5
```

**Root cause (observed):** `MarkdownFileChunker` only chunks the Markdown body. The frontmatter fields (`name`, `description`, `tags`) are parsed into metadata but are not emitted as searchable text, so the BM25 keyword index never sees the Chinese description.

### Bug 2: `stat`/`frontmatter_read`/`frontmatter_update` require `.md` suffix while `edit` does not

In the "Files and Daily Notes" section, the docs show commands without the `.md` suffix:

```bash
reme stat path=digest/wiki/quick-start-demo
reme frontmatter_read path=digest/wiki/quick-start-demo
reme frontmatter_update path=digest/wiki/quick-start-demo metadata='{"tags":["demo"]}'
```

These fail with `not found` / `not markdown` errors. Only adding `.md` makes them work:

```bash
reme stat path=digest/wiki/quick-start-demo.md
reme frontmatter_read path=digest/wiki/quick-start-demo.md
```

At the same time, the `edit` example in the same section works without the suffix:

```bash
reme edit path=digest/wiki/quick-start-demo old="indexes" new="continuously indexes"
```

**Root cause (observed):** `edit` calls `gate_md()` to auto-append `.md`, but `stat`, `frontmatter_read`, and `frontmatter_update` do not.

## Expected behavior

1. `reme search query="快速开始 示例 记忆"` should return the note whose frontmatter contains that Chinese description.
2. `stat`, `frontmatter_read`, and `frontmatter_update` should behave consistently with `edit` and accept paths without an explicit `.md` suffix.

## Environment

- OS: macOS 14.6.1
- Python: 3.11
- ReMe: latest `main` at the time of testing

## Suggested fix

[...]

PR #430, opened 2026-08-06
fix(file_io): auto appending suffix for all related steps

imrewce:
fixing issue#429 by importing gate_md for frontmatter read update and file_io stat.

### DIFF
--- a/reme/steps/file_io/_path.py
+++ b/reme/steps/file_io/_path.py
@@ -66,6 +66,17 @@ def is_relative_to(path: Path, parent: Path) -> bool:
         return False
 
 
+def display_path(workspace_path: Path, target: Path) -> str:
+    """Return ``target`` as a workspace-relative POSIX path for user-facing messages.
+
+    Falls back to the absolute string when ``target`` is not under ``workspace_path``.
+    """
+    try:
+        return target.relative_to(workspace_path).as_posix()
+    except ValueError:
+        return str(target)
+
+
 # pylint: disable=too-many-return-statements
 def resolve_path(
     workspace_path: Path,
--- a/reme/steps/file_io/frontmatter_delete.py
+++ b/reme/steps/file_io/frontmatter_delete.py
@@ -14,7 +14,7 @@ from pathlib import Path
 import frontmatter
 
 from ._file_io import get_path_lock
-from ._path import resolve_path
+from ._path import display_path, gate_md, resolve_path
 from ..base_step import BaseStep
 from ...components import R
 
@@ -37,11 +37,13 @@ class FrontmatterDeleteStep(BaseStep):
         if err or target is None:
             payload: dict = {"path": path, "error": err or "invalid path"}
         else:
+            original_target = target
+            target, is_md = gate_md(target)
             lock = await get_path_lock(target)
             async with lock:
                 if not target.is_file():
-                    payload = {"path": path, "error": "not found"}
-                elif target.suffix != ".md":
+                    payload = {"path": path, "error": f"{display_path(workspace_dir, target)} not found"}
+                elif not is_md:
                     payload = {"path": path, "error": "not markdown"}
                 elif not keys:
                     payload = {"path": path, "error": "keys is empty"}
@@ -63,6 +65,8 @@ class FrontmatterDeleteStep(BaseStep):
                         "missing": missing,
                         "frontmatter": dict(post.metadata),
                     }
+            if target != original_target:
+                payload["resolved_path"] = display_path(workspace_dir, target)
 
         if "error" in payload:
             self.context.response.success = False
--- a/reme/steps/file_io/frontmatter_read.py
+++ b/reme/steps/file_io/frontmatter_read.py
@@ -13,7 +13,7 @@ from pathlib import Path
 import frontmatter
 import yaml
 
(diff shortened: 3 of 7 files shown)


jinliyl (commented):
I checked out the branch and diffed actual behavior against `main` before writing this — repro details are inline below. There's one real regression and one missed step; the rest are smaller.

### 1. `stat` regresses directory probing

The new fallback in `stat.py` only kicks in when the `.md` candidate does **not** exist:

```python
if not target.exists() and original_target != target and original_target.exists():
```

But in this repo's daily layout, `<daily_dir>/<date>/` (a directory) and `<daily_dir>/<date>.md` (the index written by `refresh_day_index`, `_daily_index.py:97`) always coexist. So the `.md` branch wins and the directory is never reported:

| `stat path=daily/2026-08-10` | main | this PR |
|---|---|---|
| `type` | `dir` | **`file`** |
| `absolute_path` | `…/2026-08-10` | **`…/2026-08-10.md`** |
| `mime` | `None` | `text/markdown` |

`stat.py`'s own docstring states `type` is `"file"` / `"dir"` and that it "covers event workspace probes too" — directory probing is a supported use of this step. `test_stat_directory_fallback` only seeds a directory with no same-named `.md`, which is why CI stays green.

Suggested fix — give an existing directory top priority:

```python
target, _is_md = gate_md(target)
if original_target.is_dir() or (not target.exists() and original_target.exists()):
    target = original_target
```

### 2. `stat` silently shadows extension-less files, and the payload contradicts itself

With both `LICENSE` (15 bytes) and `LICENSE.md` (12 bytes) present, `stat path=LICENSE` now reports `size=12`, `mime=text/markdown`, `absolute_path=…/LICENSE.md` — while `payload["path"]` is still `LICENSE`. The two fields disagree, and nothing lets the caller notice the substitution.

`write.py` appends `[system notice: …]` when it operates in a degraded/unexpected mode. Something equivalent would help here — e.g. surface a `resolved_path` in the metadata whenever the suffix was actually appended.

### 3. `frontmatter_delete` was missed, so the family is now inconsistent

The title says "all related steps", but `frontmatter_delete.py:44` still has `target.suffix != ".md"`. Same suffix-less path, run against this branch:

```
frontmatter_read   → success=True
frontmatter_update → success=True
frontmatter_delete → success=False   Error: not found
```

Three sibling ops disagreeing on path handling is arguably a worse trap than the original bug.

### 4. Scope wording in the description

The body says "fixing issue#429", but this only covers Bug 2 of the two reported there. Not a blocker, and the literal string won't auto-close the issue (GitHub only reacts to `fix`/`fixes`/`fixed`/`closes`/… immediately followed by `#429`), plus @ployts has already scheduled Bug 1 (frontmatter indexing) as a separate follow-up. Still worth stating the split explicitly so the issue doesn't look fully resolved.

### 5. Errors hide the path that was actually probed

`frontmatter_read path=digest/missing` reports `Error: digest/missing not found`, but the path actually checked on disk was `digest/missing.md`. Neither the answer nor the metadata ever mentions the appended suffix, which makes this harder to debug than it needs to be. Same in `frontmatter_update`.

### Two pre-existing limits, not introduced here

- `gate_md` can't help dotted filenames: `notes/v1.2` has suffix `.2`, so it's classified non-markdown and never gets `.md` appended. If any doc example uses such a path, #429's symptom resurfaces in a different shape.
- `frontmatter_read` and `stat` don't check `_allowed_paths`, while `read` / `load` / `save` / `write` / `edit` all do. Out of scope here, but probably worth its own issue.

### What's right

Moving `_check_path_permission` after `gate_md` in `frontmatter_update` is a necessary correctness fix — the old order validated the pre-suffix path — and the new ordering matches `write.py:48-50` and `edit.py:54-56`. Placing `gate_md` before the `is_file()` check in `frontmatter_read` also keeps the non-markdown semantics equivalent to before, with no new hole.

imrewce:
Below are the corresponding further adjustments according to previous CR comments, to address the potential risk of misuse in file_io:
- `reme/steps/file_io/_path.py` — Added a `display_path()` helper to output workspace-relative paths for use in messages and metadata.

- `reme/steps/file_io/stat.py:57` —
  - **Comment 1**: Updated logic to `original_target.is_dir() or (not target.exists() and original_target.exists())`. This ensures that existing directories take precedence (addressing the daily layout scenario where `<date>/` and `<date>.md` coexist), correctly reporting `type="dir"`.
  - **Comments 2 & 5**: When suffix substitution occurs, `resolved_path` is added to the payload. For "not found" errors, both the answer and metadata now report the actual probed `.md` path.

- `reme/steps/file_io/frontmatter_delete.py:41` —
  - **Comment 3**: Added `gate_md()` call to align behavior with `read`/`update` operations. The lock is now applied to the path after suffix addition.
  - Includes `resolved_path` in the response and provides error messages indicating the actual probed path.

- `reme/steps/file_io/frontmatter_read.py:38` / `frontmatter_update.py:47` —
  - **Comment 5**: Updated "not found" errors to explicitly state the probed path (e.g., `Error: digest/missing.md not found`).
  - Metadata now includes `resolved_path`.
  - On success, `resolved_path` is exposed when suffix substitution occurs.

jinliyl (approved):
LGTM

PR #452, opened 2026-08-12
fix(docs): make homepage cards use direct links

jinliyl:
## Summary

- make the documentation homepage cards use direct `?doc=` URLs instead of asynchronous delegated navigation
- apply the same reliable navigation to the homepage quick-start and project links
- rename the benchmark call to action so it accurately describes its BEAM destination

## Validation

- `cd github-pages && npm run build`
- `node --check github-pages/src/main.js`
- `git diff --check`


ployts (approved):
LGTM

PR #499, opened 2026-08-27
ci: allow bootstrap release before qwenpaw plugins

jinliyl:
## Summary

- add an explicit switch for verifying published qwenpaw plugin dependencies
- keep verification enabled by default
- allow bootstrap releases to publish the reme-ai version required by independently versioned plugins

## Context

The release workflow currently deadlocks when new plugin versions require the unreleased reme-ai version, while the reme-ai release requires those plugin versions to already exist on PyPI. Bootstrap releases can now disable only the qwenpaw online installation check, publish reme-ai, and then publish the plugins.

## Validation

- python -m pytest tests/unit/test_package_versions.py -q
- pre-commit run --files .github/workflows/_build-python-packages.yml .github/workflows/release-python.yml
- git diff --check

### DIFF
--- a/.github/workflows/_build-python-packages.yml
+++ b/.github/workflows/_build-python-packages.yml
@@ -13,6 +13,11 @@ on:
         required: false
         default: false
         type: boolean
+      verify_qwenpaw_dependencies:
+        description: Verify that independently published qwenpaw plugins are installable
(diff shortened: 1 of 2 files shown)


ployts (approved):
LGTM

PR #502, opened 2026-08-27
ci: update release workflow actions and smoke checks

jinliyl:
## Summary

- upgrade pinned upload-artifact references from v4 to v6
- upgrade pinned download-artifact references from v4 to v7
- update all Python, plugin, Studio, and TypeScript release workflows consistently
- migrate the Auto Fin wheel smoke test from legacy Python plugin/config entry points to the current package-only plugin.yaml contract

## Context

GitHub-hosted runners warn that the pinned v4 artifact actions target deprecated Node.js 20 and are being forced onto Node.js 24. The updated versions declare Node.js 24 directly while preserving the repository SHA-pinning policy.

The Auto Fin release smoke test also expected the old Plugin descriptor and reme.configs entry point. Auto Fin now exposes a package-only reme.plugins entry point and declares its backends and jobs in plugin.yaml, so the release check now validates that current contract like the Daily Paper workflow.

## Validation

- pre-commit checks for all changed workflows
- python -m pytest plugins/auto-fin -q (9 passed)
- built the Auto Fin wheel, installed it into an isolated target, and verified its entry point and packaged manifest
- git diff --check
- verified no v4 upload-artifact or download-artifact references remain under .github/workflows

### DIFF
--- a/.github/workflows/_build-python-packages.yml
+++ b/.github/workflows/_build-python-packages.yml
@@ -102,7 +102,7 @@ jobs:
 
       - name: Upload ReMe distributions
         if: inputs.upload_artifacts
-        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
+        uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6
         with:
           name: reme-distributions
           path: dist/reme/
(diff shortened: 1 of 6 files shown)


ployts (approved):
审查结论：未发现需要阻塞合并的问题，建议合并。

已核验：
- upload-artifact v6 与 download-artifact v7 的固定 SHA 均对应官方版本，且使用 Node.js 24；所有相关工作流引用已一致更新。
- Auto Fin wheel 可成功构建，隔离安装后的插件入口与 plugin.yaml 清单检查通过。
- Auto Fin 测试 9 项全部通过，git diff --check 通过，PR 当前全部检查均为绿色。

剩余风险：发布工作流为手动触发且依赖发布环境与凭据，本次未进行端到端发布验证。

PR #517, opened 2026-09-04
feat(Tag filter): Add an optional rebuildable tag index for file frontmatter

xyf2020:
## Summary

Add an optional, in-memory tag index derived from file frontmatter. Workspace files and the file graph remain authoritative; the tag index is rebuildable derived state and introduces no persistence file or dedicated watcher.

Key behavior:

- Adds the `tag_index` component type and a local bidirectional path/tag backend.
- Reads tags from `FileNode.front_matter` and normalizes them for deterministic AND/OR lookup.
- Updates tags through the existing `LocalFileStore` mutation lifecycle and rebuilds them from the file graph on startup.
- Keeps tag-index failures non-blocking for graph, chunk, and keyword-index mutations.
- Quarantines an index after failed reconciliation so queries fail closed instead of returning stale matches.
- Extends the existing `reindex` job with `scope=tag`; `scope=all` now rebuilds BM25, embeddings, and tags without rescanning workspace files.
- Adds optional AutoMemory tag generation and normalization through `enable_tags`. It is disabled in the default configuration.

Public search filtering by tag is intentionally left out of scope. This PR provides the component, lifecycle integration, lookup primitives, and rebuild contract needed for that follow-up.

## Related issue

N/A

## Contract and data impact

- [ ] No public configuration, schema, CLI, endpoint, streaming, or workspace-layout contract changes
- [x] No user-owned memory files are deleted or rewritten
- [x] Derived indexes, catalogs, graphs, caches, and metadata remain rebuildable

This PR adds an optional `tag_index` component configuration and extends the public `reindex` scope enum with `tag`. The default file store does not enable the tag index; commented configuration shows how to opt in.

AutoMemory understands the `enable_tags` step option, but the default job explicitly sets it to `false`. Enabling it allows newly created or updated memory notes to receive normalized frontmatter tags. There is no bulk migration of existing files.

## Validation

- [x] Focused tests pass
- [ ] Unit tests pass, or omitted/failing tests are explained below
- [ ] `pre-commit run --all-files` passes, or omitted checks are explained below
- [x] Frontend checks were run when `reme_studio/` changed

Checks run for the latest update:

- `python -m pytest tests/unit/test_tag_index.py tests/unit/test_reindex.py tests/unit/test_file_store_consistency.py -q`
  - 99 passed
- `python -m pytest tests/unit -q`
  - 1148 passed, 1 failed
  - The remaining failure is `test_stdlib_forwards_screen_and_file_logs_to_qwenpaw`; it also fails in isolation and does not exercise any files changed by this PR.
- `git diff --check`
  - Passed

`pre-commit run --all-files` was not run. No frontend checks were required because `reme_studio/` was not changed. Integration tests were not run because this change does not require external services or credentials.

## Checklist

- [x] I reviewed the diff for unrelated changes and sensitive data
- [x] Tests cover tag normalization, mutation synchronization, failure recovery, fail-closed lookup, startup rebuild, and explicit reindexing
- [x] Defaults and job schemas were updated with the implementation
- [x] The tag index remains optional, in-memory, and managed by the existing component lifecycle

## Screenshots or additional notes

The tag index does not introduce a background task, separate parser, persistence file, or additional long-lived resource. It consumes the existing graph snapshot and watcher-driven file-store updates.


jinliyl (approved):
LGTM

PR #520, opened 2026-09-04
docs: refresh ReMe homepage branding

jinliyl:
## Summary

- introduce a ReMe wordmark and glass-style app icon for the documentation homepage
- refine the hero artwork while preserving the bilingual left/right layout
- give homepage feature cards distinct accent colors and align each icon with its title
- copy and verify the new logo asset as part of the disposable GitHub Pages build

## Validation

- `npm test` in `github-pages` (7 tests passed)
- `npm run build` in `github-pages` (19 artifacts verified)
- `pre-commit run --files ...` for all changed files
- `git diff --check`

PR #521, opened 2026-09-04
fix(docs): use uncached navigation icon

jinliyl:
## Summary

Follow-up to #520.

- use the new glass ReMe icon through a dedicated `/reme-icon.svg` asset
- point both the document favicon and navigation brand at the new asset URL
- remove the obsolete `docs/public/favicon.svg` path to avoid stale browser and CDN caches
- update generated-content tests and build artifact verification

## Validation

- `npm test` in `github-pages` (7 tests passed)
- `npm run build` in `github-pages` (19 artifacts verified)
- `git diff --check`

### DIFF
--- a/docs/.vitepress/config.mts
+++ b/docs/.vitepress/config.mts
@@ -256,7 +256,7 @@ export default defineConfig({
     },
   },
   head: [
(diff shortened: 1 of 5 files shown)


ployts (approved):
LGTM

PR #525, opened 2026-09-06
fix(embedding): retry 429 rate-limit errors instead of dropping the batch

### LINKED ISSUE #523: [Bug]: Embedding backfill/rebuild dies instantly on 429 rate-limit (no retry/backoff) and leaves vector search permanently disabled after failure
## Observed

During an embedding index rebuild (or startup backfill), any batch embedding request that
returns HTTP 429 is NOT retried. `LocalEmbeddingStore._call_with_retry` only retries
`TimeoutError` / `ConnectionError` / `OSError` with exponential backoff (`2**attempt`).
`openai.RateLimitError` is not a subclass of those (MRO: Exception -> APIError ->
APIStatusError), so a 429 falls into the generic `except Exception` branch, where only
errors with code `insufficient_quota` are retried (and only when `quota_retry_delay` is
configured, which is None by default). Every other status error logs
`Embedding request failed` and returns `None` immediately — the whole batch dies, zero
retries, zero backoff.

Consequences once the rebuild is triggered while the provider is rate-limited:

1. The reindex job ends with `RuntimeError: embedding reindex incomplete: N chunks failed`.
2. The `needs_reindex` gate stays active (by design of #508: "failure or cancellation keeps
   the gate active so the operation can be retried safely"), but nothing retries it.
3. On every subsequent process restart the startup backfill is skipped with
   `reason=manual_reindex_required`, and `_embedding_rebuild_pending` makes
   `_get_query_embedding` return `None`, so **vector search returns zero hits** (BM25
   keyword search remains the only working retrieval).
4. The user must manually re-trigger the rebuild, which then fails again on the first 429.

Net effect: an embedding model can become effectively permanently disabled after a single
rate-limit burst, with no automatic recovery.

## Expected

- HTTP 429 (and probably transient 5xx) status errors on the batch embedding path should be
  retried with exponential backoff (+ jitter), like `TimeoutError`/`ConnectionError`/`OSError`
  already are.
- Alternatively/additionally, a failed rebuild should automatically retry the missing chunks
  in the background (or clear the gate and let the normal backfill fill missing vectors once
  the provider recovers), so a rate-limit burst does not permanently disable vector search.

### Steps to reproduce

1. Configure an embedding provider that rate-limits (e.g. an OpenAI-compatible gateway that
   returns 429 `AccountRateLimitExceeded` after a burst). Config example (values redacted):
   backend=openai, base_url=<gateway>, model_name=<embedding model>, dimensions=2048,
   max_batch_size=10.
2. Trigger a full embedding reindex (switch to a new embedding model so the vector space
   changes and every chunk is missing, then run the reindex job with scope=embedding).
3. Observe: the first batch that gets a 429 logs `Embedding request failed` and is dropped
   immediately; every subsequent 429 batch does the same.
4. The job ends with `embedding reindex incomplete: N chunks failed`.
5. Restart the process: startup backfill is skipped (`reason=manual_reindex_required`),
   and vector search returns 0 hits while the gate stays pending until a manual re-click —
   which fails again on the first 429.

### Relevant configuration

```yaml
- embedding backend: openai-compatible (code path shared by OpenAI-compatible / DashScope /
  Gemini / Ollama backends)
- max_batch_size: 10
- max_retries: 3
- quota_retry_delay: not set (None)
- health_check: passes OK at startup; failure is on the batch path, not the health probe
```

### Logs or traceback

```shell
# repeated on every rate-limited batch:
ERROR reme.components.embedding_store.local_embedding_store | Embedding request failed
Traceback (most recent call last):
  ...
  File ".../agentscope/embedding/_openai/_model.py", line 152, in _call_api
    response = await self.client.embeddings.create(**api_kwargs)
  ...
openai.RateLimitError: Error code: 429 - {'error': {'code': 'AccountRateLimitExceeded',
  'message': 'Requests are too frequent. Please reduce your request frequency, ...'}}

# end of the rebuild:
ERROR reme.components.job.base_job | Failed to execute job: embedding reindex incomplete: 37 chunks failed

# after restart, backfill permanently skipped and vector search disabled:
INFO reme.components.file_store.local_file_store | default: embedding backfill skipped: reason=manual_reindex_required
INFO reme.steps.index.search | [SearchStep] query='...' candidates=6 vector_hits=0 keyword_hits=6
```

### ReMe version

0.4.1.10

### Python version

3.11.9

### Operating system

Linux

### Affected area

Other

### Data safety

- [x] I removed credentials and private memory content from this report.

AmirF194:
Added `_is_rate_limited` right next to `_is_insufficient_quota`, built the same way (checks `status_code` then a body-embedded code, no SDK import), and route a 429 through the same exponential backoff `TimeoutError`/`ConnectionError`/`OSError` already get. It isn't gated behind `quota_retry_delay` like the quota branch: a rate limit is transient by definition, so it always retries now.

I scoped this to 429 only. You also mentioned "probably transient 5xx" and backoff jitter; I didn't reproduce either being needed here and didn't want to guess past what the issue actually shows, so both are left out.

Fixes #523.

### DIFF
--- a/reme/components/embedding_store/local_embedding_store.py
+++ b/reme/components/embedding_store/local_embedding_store.py
@@ -194,6 +194,10 @@ class LocalEmbeddingStore(BaseEmbeddingStore):
                 if attempt < self.max_retries - 1:
                     await asyncio.sleep(2**attempt)
             except Exception as error:
+                if self._is_rate_limited(error):
+                    if attempt < self.max_retries - 1:
+                        await asyncio.sleep(2**attempt)
+                    continue
                 if (
                     self.quota_retry_delay is not None
                     and self._is_insufficient_quota(error)
@@ -209,6 +213,17 @@ class LocalEmbeddingStore(BaseEmbeddingStore):
                 return None
         self.is_healthy = False
         return None
+
+    @staticmethod
+    def _is_rate_limited(error: Exception) -> bool:
+        """Recognize an OpenAI-compatible 429 response without importing a provider SDK."""
+        if getattr(error, "status_code", None) == 429:
+            return True
+        body = getattr(error, "body", None)
+        if not isinstance(body, dict):
+            return False
+        details = body.get("error", body)
+        return isinstance(details, dict) and details.get("code") == "rate_limit_exceeded"
 
     @staticmethod
     def _is_insufficient_quota(error: Exception) -> bool:
--- a/reme/components/embedding_store/local_embedding_store.py
+++ b/reme/components/embedding_store/local_embedding_store.py
@@ -196,8 +196,13 @@ class LocalEmbeddingStore(BaseEmbeddingStore):
             except Exception as error:
                 if self._is_rate_limited(error):
                     if attempt < self.max_retries - 1:
-                        await asyncio.sleep(2**attempt)
-                    continue
+                        delay = 2**attempt
+                        self.logger.warning(f"Embedding rate limited; retrying in {delay:.1f}s")
+                        await asyncio.sleep(delay)
+                        continue
+                    self.logger.exception("Embedding request failed after exhausting rate-limit retries")
+                    self.is_healthy = False
+                    return None
                 if (
                     self.quota_retry_delay is not None
                     and self._is_insufficient_quota(error)
--- a/reme/components/embedding_store/local_embedding_store.py
+++ b/reme/components/embedding_store/local_embedding_store.py
@@ -217,6 +217,8 @@ class LocalEmbeddingStore(BaseEmbeddingStore):
     @staticmethod
     def _is_rate_limited(error: Exception) -> bool:
         """Recognize an OpenAI-compatible 429 response without importing a provider SDK."""
+        if LocalEmbeddingStore._is_insufficient_quota(error):
+            return False
         if getattr(error, "status_code", None) == 429:
             return True
         body = getattr(error, "body", None)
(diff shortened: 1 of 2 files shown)


jinliyl:
Thank you for the contribution! Retrying transient rate-limit errors with bounded exponential backoff makes sense and addresses the immediate batch failure described in #523.

I found one regression that would be good to address before merging: an OpenAI-compatible insufficient_quota error can also have status_code=429. Since the new rate-limit branch runs first, it bypasses the existing quota handling. I reproduced this using an openai.RateLimitError: quota_retry_delay=60 waits only 1 second, and quota_retry_delay=None now retries instead of returning immediately.

Could you exclude errors recognized by _is_insufficient_quota from the generic rate-limit branch, preserving the existing quota behavior? A regression test with both status_code=429 and code=insufficient_quota would help cover this; the current quota test double does not include the HTTP status code.

The 28 focused embedding tests passed on this branch. Thanks again for tackling this issue!

AmirF194:
Good catch, fixed. _is_rate_limited now checks _is_insufficient_quota first and defers to it, so an insufficient_quota error with status_code=429 no longer takes the generic branch.

Reproduced your case first: set status_code = 429 on the existing InsufficientQuotaError test double (it only had the body code before), which makes both quota tests fail without the fix and pass with it. Added a direct unit test on the classifier too. All 28 tests pass in a clean container.

jinliyl (approved):
LGTM

PR #537, opened 2026-09-11
fix(daily-paper): keep digest generation after empty search

jinliyl:
## Summary

- prioritize generating the daily paper brief from the three supplied analyses
- limit historical-memory search to two calls and treat empty results as no related history
- keep wikilink discovery optional without misclassifying the paper inputs as missing
- update prompt assertions to cover the fallback behavior

## Testing

- python -m pytest plugins/daily_paper/tests/test_daily_paper.py -v
- pre-commit run --files plugins/daily_paper/src/reme_daily_paper/digest.yaml plugins/daily_paper/tests/test_daily_paper.py


ployts (approved):
LGTM

PR #542, opened 2026-09-14
chore(deps): bump CodeQL action to 4.38.0

jinliyl:
## Summary

- update both `github/codeql-action/init` and `github/codeql-action/analyze` to 4.38.0
- group CodeQL actions in Dependabot so coupled entrypoints are updated together

This supersedes #540 and #541, which update only one CodeQL entrypoint at a time and fail because `init` and `analyze` require matching action versions.

## Validation

- `git diff --check`
- parsed both changed YAML files with PyYAML


ployts (approved):
LGTM

PR #544, opened 2026-09-14
fix(job): preserve nested exception details

jinliyl:
## Summary

- keep exception class names when an exception has an empty string representation
- include explicit causes and unsuppressed context in captured job errors
- guard exception-chain traversal against cycles

This lets hosts surface the underlying connection-reset detail for the failure reported in agentscope-ai/QwenPaw#7715.

## Testing

- `pytest tests/unit/test_job.py -q`
- `pre-commit run --files reme/components/job/base_job.py tests/unit/test_job.py`


ployts (approved):
LGTM

PR #550, opened 2026-09-15
refactor(plugins): decouple DingTalk notifications

jinliyl:
## Summary

- keep Auto Fin and Daily Paper focused on producing local memory artifacts
- remove the DingTalk send step and credentials from Daily Paper defaults
- remove notifier-specific docs and tests from both producer plugins
- leave notification composition to external config and the standalone DingTalk plugin

## Validation

- `pytest plugins/auto-fin/tests plugins/daily_paper/tests -q` (42 passed)


ployts (approved):
LGTM

PR #552, opened 2026-09-15
docs(plugins): use local install paths

jinliyl:
## Summary

- install DingTalk, Auto Fin, and Daily Paper from their repository-local plugin directories
- update cookbook, framework, CLI reference, and plugin README examples in both languages
- replace the unpublished Auto Fin package name in generic package-management examples with `your-plugin-package`

This is the documentation-only follow-up to merged PR #551.

## Validation

- `python -m pytest tests/unit/test_cookbook_config.py -q` (7 passed)
- documentation tests and build previously passed against the identical final tree (13 tests, 25 artifacts verified)


ployts (approved):
LGTM

PR #557, opened 2026-09-16
fix(chunker): restore malformed JSON text fallback

### LINKED ISSUE #556: [Bug]: JSON text fallback raises TypeError when malformed input exceeds chunk_chars
### Description

`JsonFileChunker.chunk()` has an explicit text fallback for files that raise `json.JSONDecodeError`. With the default `chunk_chars=2000`, this fallback works for a 2,000-character malformed file but raises `TypeError` at 2,001 characters. Valid JSON above the limit still chunks successfully.

The large-file fallback calls `raw_text.split("\n", keepends=True)`. `str.split()` does not accept `keepends`.

Expected: syntactically malformed, decodable JSON should use the existing text fallback at either size, preserving the text and meaningful line ranges without rewriting the source file. This does not request more permissive character decoding or changes to valid JSON chunking.

For applications explicitly indexing JSON through `update_index_step`, a newly added affected file is omitted from the derived index. If an indexed file becomes malformed, the update fails and its previous chunks remain. Other files in the batch continue to index. Repairing the file to valid JSON allows its next update to succeed.

The default daily/digest index watcher watches Markdown only. The indexing impact above was reproduced with an explicit JSON indexing Job using real local BM25 and graph components, without model calls.

### Steps to reproduce

From a checkout of the commit below, install the package with its AgentScope extra, then run:

```bash
python -m pip install -e ".[as]"
python - <<'PY'
import asyncio
import tempfile
from pathlib import Path
from reme.components.file_chunker import JsonFileChunker

async def main():
    with tempfile.TemporaryDirectory() as directory:
        path = Path(directory) / "memory.json"
        path.write_text('{"note": "' + "unfinished memory entry\n" * 100, encoding="utf-8")
        await JsonFileChunker().chunk(path)

asyncio.run(main())
PY
```

This creates a 2,410-character UTF-8 file with an incomplete JSON string.

### Relevant configuration

```yaml
backend: json
encoding: utf-8
chunk_chars: 2000
```

### Logs or traceback

```text
reme/components/file_chunker/json_file_chunker.py, line 140, in chunk
    lines = raw_text.split("\n", keepends=True)
TypeError: 'keepends' is an invalid keyword argument for split()
```

Boundary checks on the unchanged upstream commit:

```text
empty: 0 chars, 0 chunks
malformed: 2000 chars, 1 chunk
malformed: 2001 chars, TypeError
malformed multiline: 2410 chars, TypeError
valid JSON: 2511 chars, 2 chunks
```

The existing JSON, JSONL, plain-text, Markdown chunker and background-step suites pass: `200 passed in 1.60s`. The existing `test_malformed_fallback` covers only a short malformed input.

### ReMe version

0.4.1.13, upstream `main` commit `5231f3970c550bb0bc5d1bc7fbc85534274ca530`, checked on 2026-09-16 UTC.

### Python version

3.12.10

### Operating system

macOS 27.0, arm64

### Affected area

Search, catalog, graph, or index

### Possible implementation and scope

Correct the line splitting in the shared malformed-JSON fallback and add focused regression coverage for the size boundary, multiline text and JSON indexing. Keep valid JSON parsing, strict encoding, source files and the default watcher configuration unchanged. No new dependency is needed.

### Data safety

- [x] I removed credentials and private memory content from this report.

baba9811:
## Summary

Malformed JSON larger than `chunk_chars` raises `TypeError` because the text fallback passes `keepends` to `str.split()`. This also prevents JSON index updates from adding new chunks or replacing previously indexed content.

Iterate the decoded text with `StringIO` to preserve LF line endings, and avoid creating an empty chunk before an oversized line. Add regressions for the default size boundary, a custom limit, CRLF normalization, terminal newlines, Unicode separators, and add/replace indexing with the real local file store.

## Related issue

Fixes #556

## Contract and data impact

- Valid JSON parsing, strict character decoding, and public configuration and schemas are unchanged.
- User-owned source files are not rewritten. Tests verify their bytes remain intact.
- Derived indexes remain rebuildable. The replacement regression verifies that old chunks are removed and a healthy sibling remains indexed.

## Validation

Python 3.12.10 on macOS arm64:

- New regressions failed on the original implementation: 5 failed, 123 passed.
- `python -m pytest tests/unit/test_json_file_chunker.py tests/unit/test_background_steps.py -q`: 128 passed.
- `python -m coverage run -m pytest tests/unit plugins/auto-fin plugins/daily_paper plugins/lme plugins/beam -v --tb=long -s --log-cli-level=WARNING`: 1,478 passed, 1 skipped. The skip requires optional `pillow_heif`; the run also emitted a Starlette/AnyIO deprecation warning.
- `pre-commit run --all-files`: passed.
- The original public `JsonFileChunker.chunk()` and `ReMe.run_job()` reproduction now succeeds for large malformed inputs, including adding and replacing index entries.

Credentialed model/service integration tests were not run because no provider environment was supplied for this change. The local chunker, Job, and indexing checks above use no model calls. The Linux/Python CI matrix has not been run locally.


xyf2020:
Non-blocking thought: instead of a bespoke fallback here, how about delegating
to the default chunker on JSONDecodeError (wrap json.loads in try/except and
fall through)? This path had been silently broken since #325, so piggybacking
on the primary text chunker would keep it covered by well-tested code and let
us drop this custom logic.

One caveat: the default chunker is byte-based with overlap, while this fallback
is line-aligned without overlap — not strictly equivalent, and the delegation
would likely live in UpdateIndexStep (which owns chunker selection) rather than
here.

baba9811:
@xyf2020 I agree that reusing the default chunker would reduce duplicate fallback logic. I'd prefer to keep this PR focused on fixing the crash while retaining the existing line-aligned, non-overlapping approach, given the character/byte sizing and overlap differences you noted.

Moving the fallback into `UpdateIndexStep` would also require deciding what happens for direct `JsonFileChunker.chunk()` calls and configurations that register only a JSON chunker. The regressions here cover both paths without requiring a default chunker.

Could we keep this fix scoped as-is and consider delegation in a separate follow-up, with those behaviors made explicit?

PR #562, opened 2026-09-21
feat(scripts): plot cumulative star growth over a configurable period

jinliyl:
## Summary

- rebuild the daily cumulative star total from the repository's current star count and the daily increments, so the script can plot where the repository stands rather than only how many stars arrived
- render the chart as an SVG line chart (with an area fill and the final total annotated) instead of a bar chart of daily increments
- add `--period` to control the time window, accepting `d`/`w`/`m`/`y` units and defaulting to `90d`; the exact-day units count inclusively and the calendar units anchor to the same day of the month
- keep `--days N` as an alias for `--period N`, and reject passing both
- crop the vertical axis to the neighbourhood of the data instead of anchoring it at zero, so a few hundred new stars stay visible next to a few thousand existing ones; data that starts near zero still plots from zero
- label the chart with the plotted date range and the final star count, and report the star count at the start of the window on stderr

## Validation

- `python -m black --line-length=120 --target-version=py311 --check scripts/github_star_growth.py`
- `python -m flake8 --extend-ignore=E203 --max-line-length=120 scripts/github_star_growth.py`
- live run against `agentscope-ai/ReMe` through `gh`, with the default window and with `--days 180`
- offline geometry checks of the axis logic on synthetic data (flat series, zero-based series, single point) to confirm the curve fills the plot and never divides by zero
- `GH_TOKEN`-free error paths: `--period 0d`, `--period abc`, and `--period` combined with `--days` all fail with a clear message

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ployts (approved):
LGTM

Issue #379, opened 2026-07-20
fix: make metadata-backed step answers complete for LLM consumers


jinliyl:
## Context

ReMe MCP tools are designed for direct LLM consumption and intentionally expose Response.answer. Response.metadata remains auxiliary request data for programmatic clients and diagnostics.

Some Steps store information required for the next LLM action only in metadata, leaving answer incomplete. For example, list returns only a count in answer while file paths are available only in metadata, so an agent cannot select a file for a subsequent read call.

This issue follows up on #376 at the Step output layer rather than changing the MCP transport contract.

## Scope

Audit and, where needed, improve the LLM-facing answer produced by these Steps:

- list: include the returned workspace-relative file paths, not only the count.
- search: keep result paths, source text, and relevant link expansion self-contained; return an explicit message when there are no results.
- daily_list: ensure the date, count, note paths, and relevant frontmatter are clear in the answer.
- traverse: ensure the traversal summary and actionable edge details are present in the answer.
- node_search: ensure hit paths and the fields needed for follow-up reads are present in the answer.
- proactive: include the topics and requested content, rather than only a read summary; keep skipped and error outcomes explicit.

Do not copy metadata wholesale into answer. Format only information needed by an LLM to understand the result and perform the next tool call.

If other Steps have the same problem, please add concrete examples to this issue so they can be included in the audit.

## Acceptance criteria

- Each affected answer is self-contained and actionable for direct LLM consumers.
- Existing structured metadata remains available to HTTP and other programmatic consumers.
- MCPService continues to return Response.answer directly.
- Empty, skipped, and error outcomes remain concise and unambiguous.
- Focused unit tests assert the relevant LLM-visible answer content for each changed Step.

PR #563, opened 2026-09-21
fix(traverse): return an LLM-readable answer instead of a raw graph dump

RerankerGuo:
## Summary

`traverse` is exposed as an MCP/chat read-only tool (it is listed in `ChatStep.READ_ONLY_TOOLS` and the bundled `reme-memory` skill tells agents to call it), and `MCPService` hands `response.answer` straight to the LLM. The step assigned `graph.model_dump()` to that answer, so the model received a nested `{"version", "seeds", "depth", "direction", "nodes", "edges"}` dict with no summary: it had to infer hop structure itself and could not tell an indexed file from a dangling wikilink target before issuing a follow-up `read`.

This is the `traverse` bullet of #379 ("ensure the traversal summary and actionable edge details are present in the answer").

## Change

Render the answer as a self-contained, LLM-actionable summary:

```
=== traverse seeds=a.md depth=1 direction=forward nodes=3 edges=2 ===
[0] a.md
[1] b.md
[1] c.md
--- edges ---
a.md -> b.md (depth=1)
a.md -> c.md#intro (depth=1)
```

- one line per reached node with its hop depth; unresolved wikilink targets are marked `(unindexed)` so a caller does not try to `read` a path with no indexed node
- one line per directed edge with its anchor (`target#anchor`) and traversal depth
- an edge-free traversal states `(no wikilink edges found)` explicitly instead of returning a bare structure

The structured graph is preserved in `metadata["graph"]`, so HTTP and other programmatic consumers keep the full object; only the LLM-facing answer changes shape.

## Deliberately out of scope

`graph_snapshot` (`reme/steps/index/graph_snapshot.py:84`) and `app_config` (`reme/steps/common/app_config.py:84`) also assign object answers, but ReMe Studio reads `.answer` directly for both (`reme_studio/app/api.ts:61` and `:82`, typed as `AppConfig` / `GraphSnapshot`). Changing those would break the frontend, so they are left untouched — that part of #379 needs a coordinated Studio change.

## Validation

- `pytest tests/unit` — **1314 passed, 1 skipped** (was 1312 passed, 1 skipped before; +2 new tests)
- `pytest tests/unit/test_common_steps.py -q` — 19 passed
- `black --line-length=120 --target-version=py311 --check` — clean
- `flake8 --extend-ignore=E203 --max-line-length=120` — clean
- `pylint` with the repo's `.pre-commit-config.yaml` disable list — 10.00/10
- `add-trailing-comma` — clean
- Confirmed the two new tests fail against the unmodified `traverse.py` (checked out the original file, re-ran, restored)

New tests:

- `test_traverse_answer_is_llm_readable_and_actionable` — asserts the header, per-node depth lines, edge lines with anchors, and that `metadata["graph"]` still round-trips to `TraverseGraph`
- `test_traverse_answer_marks_unindexed_and_empty_results` — asserts the `(unindexed)` marker for a dangling target and the explicit no-edges notice

Existing traverse tests that read the graph out of `response.answer` now read it from `metadata["graph"]` via a small `_graph()` helper. No other consumer of `traverse`'s answer exists in the repo (checked `reme_studio/`, `typescript/`, `integrations/`, `plugins/`, `benchmark/`, `skills/`).

## Notes

- `Response.answer` is typed `str | Any`, so a string answer is within the existing contract, and `CliService._format_response` already prefers a string answer (it only JSON-dumps non-strings).
- No config keys, schemas, endpoint shapes, or the default watcher behaviour are changed.


skills/reme_memory/SKILL.md
```diff
@@ -150,6 +150,10 @@ Use `traverse` when wikilink neighbors may matter:
 reme traverse path="<workspace-relative-path>" depth=1 direction=both
 ```
 
+It answers with a readable summary: one line per reached node with its hop depth, then one line per directed
+edge as `source -> target#anchor (depth=N)`. A target marked `(unindexed)` has no indexed node, so do not
+`read` it.
```
ployts: [P2] 不要根据 `(unindexed)` 标记禁止读取文件

`indexed` 只表示 `file_store.get_nodes()` 中是否存在该路径对应的 FileNode，并不表示磁盘文件是否存在或能否读取。工作区中未纳入 watch_dirs 的 Markdown 文件，以及尚未完成索引的文件，都可能被标记为 `(unindexed)`，而 `ReadStep` 直接读取工作区文件，不要求它已被索引。

在本提交上已复现：索引中的 `digest/a.md` 链接到实际存在但未索引的 `notes/reference.md`；`traverse` 返回 `[1] notes/reference.md (unindexed)`，随后 `read(path="notes/reference.md")` 成功。这里的 `so do not read it` 会让使用该 skill 的代理跳过有效记忆。

建议将说明改为“该路径尚无索引记录，文件是否存在需另行确认；需要时可通过 stat/read 检查”，并同步修正 `integrations/claude_code/reme/skills/reme-memory/SKILL.md` 第 35–36 行及 `_render_answer` 中相同含义的注释。

ployts (changes requested):
已复核最新提交 `8b5bed2`。Python 实现与此前审查的 `d58df4a` 一致，未发现新的代码正确性问题；但新增的两份 skill 说明把“未索引”作为禁止读取的依据，会影响代理召回有效文件，建议修正后再通过。

验证情况：此前在 `d58df4a` 上运行的 43 项相关测试全部通过，并验证了 MCP 输出、JSON 序列化、反向遍历、锚点、多种子及 depth=0。本次在最新提交上额外复现了“traverse 标记为 unindexed，但 read 成功”的情况。未在本地运行全量单元测试、pre-commit、集成测试或文档构建。

RerankerGuo:
Thanks for the catch — you are right, and I reproduced it before fixing.

`indexed` in `TraverseGraph` only records whether the path has a node in the **file graph**, while `read` resolves against the **filesystem** (`ReadStep` -> `read_file_lines_safe`). So a workspace file that has not been indexed yet is flagged by `traverse` and still reads fine. Local repro with `a.md -> target.md`, where `target.md` exists on disk but is absent from the graph:

```
[1] target.md (unindexed)
--- edges ---
a.md -> target.md (depth=1)
```
and `read path="target.md"` returns `# Target\nreal content`.

So telling agents not to read an `(unindexed)` target would make them skip valid files. Fixed in `a1a85e33`:

- Both skill copies now say `(unindexed)` means traverse has no frontmatter or hop metadata for the path, not that the path is missing, and explicitly say it may still exist on disk.
- `_render_answer`'s docstring made the same wrong claim and is corrected.
- Added `test_traverse_unindexed_node_is_still_readable`, which seeds a real file that is absent from the graph and asserts both the marker and a successful `read`, so the claim cannot drift again.

Validation: `pytest tests/unit` = 1315 passed / 1 skipped (was 1312 at baseline), black / flake8 / pylint / add-trailing-comma clean.

Note I did not change the marker itself — it is still accurate as "no graph node", and `graph_snapshot`/`app_config` remain untouched since Studio reads `.answer` for those.

